PR #6, opened 2022-12-19
kernel: Include task_stack.h in sucompat.c

hmtheboy154:
Fixed the following error when compiling

ld.lld: error: undefined symbol: task_stack_page
>>> referenced by ld-temp.o
>>>               vmlinux.o:(sh_user_path)


tiann:
Thanks!

PR #10, opened 2022-12-20
kernel: switch to SYSCALL_WORK_SECCOMP if CONFIG_GENERIC_ENTRY is ena…

hmtheboy154:
…bled

From Kernel 5.11 it migrate to syscall_work instead of tif_*, so check for CONFIG_GENERIC_ENTRY and use SYSCALL_WORK_SECCOMP instead of TIF_SECCOMP

https://www.mail-archive.com/linux-kernel@vger.kernel.org/msg2382193.html

### DIFF
--- a/kernel/ksu.c
+++ b/kernel/ksu.c
@@ -57,7 +57,11 @@ void escape_to_root() {
 	memset(&cred->cap_ambient, 0xff, sizeof(cred->cap_ambient));
 
 	// disable seccomp
+#ifdef CONFIG_GENERIC_ENTRY
+	current_thread_info()->flags &= ~SYSCALL_WORK_SECCOMP;
+#else
 	current_thread_info()->flags &= ~TIF_SECCOMP;
+#endif
 	current->seccomp.mode = 0;
 	current->seccomp.filter = NULL;
 


kernel/ksu.c
```diff
@@ -57,7 +57,11 @@ void escape_to_root() {
 	memset(&cred->cap_ambient, 0xff, sizeof(cred->cap_ambient));
 
 	// disable seccomp
+#ifdef CONFIG_GENERIC_ENTRY
+	current_thread_info()->flags &= ~SYSCALL_WORK_SECCOMP;
```
tiann, line 61: should be `current_thread_info()->syscall_work`: https://elixir.bootlin.com/linux/v5.11.22/source/include/linux/thread_info.h#L135

tiann (changes requested):

kernel/ksu.c
hmtheboy154: I thought we can also use `SYSCALL_WORK_SECCOMP` ? I can be able to compile
https://elixir.bootlin.com/linux/v5.11.22/source/include/linux/thread_info.h#L50

tiann: I mean `current_thread_info()->syscall_work` instead of  `current_thread_info()->flags`

hmtheboy154: > should be `current_thread_info()->syscall_work`: https://elixir.bootlin.com/linux/v5.11.22/source/include/linux/thread_info.h#L135

I'll make another PR for this

PR #17, opened 2022-12-22
👷 [CI] Use checkout instead of mainline setup.sh

Ylarod:


Ylarod:
https://github.com/Ylarod/KernelSU/actions/runs/3753868441

PR #19, opened 2022-12-23
kernel: add CONFIG_KSU_DEBUG

Ylarod:


tiann:
why `shell` should be always allowed? App with wireless debug can acquire root with this feature

Ylarod:
1. _Wireless Debug_ needs to be enabled by user manually, the users using KernelSU should know the risk of giving adb permission to apps.
2. Developers can have access to `su` without installing the manager by `adb shell`

tiann:
> 1. _Wireless Debug_ needs to be enabled by user manually, the users using KernelSU should know the risk of giving adb permission to apps.
> 2. Developers can have access to `su` without installing the manager by `adb shell`

Users don't know giving adb permission would escape to root silently.

Ylarod:
if KERNEL_SU_ALLOWLIST does not exist, allow `shell` by default

tiann:
I think this should be a kernel CONFIG, if developer need this, he can open the config and build the kernel. normal users shouldn't have this feature.

Ylarod:
maybe the `ksu_config` in #61 can be used in debug mode

kernel/allowlist.c
```diff
@@ -181,7 +181,16 @@ void do_load_allow_list(struct work_struct *work)
 	fp = filp_open(KERNEL_SU_ALLOWLIST, O_RDONLY, 0);
 
 	if (IS_ERR(fp)) {
+#ifdef CONFIG_KSU_DEBUG
+		int errno = PTR_ERR(fp);
+		if (errno == -ENOENT) {
```
tiann, line 186: Do we need to check errno here?

kernel/allowlist.c
```diff
@@ -181,7 +181,16 @@ void do_load_allow_list(struct work_struct *work)
...
+			ksu_allow_uid(2000, true); // allow adb shell by default
+		}else{
```
tiann, line 188: This code is not correctly formatted.

tiann (changes requested):

kernel/allowlist.c
Ylarod: allow it only when KERNEL_SU_ALLOWLIST does not exist

tiann: seems reasonable

tiann:
Can we make the expected_size and expected_hash a kernel parameter in debug mode?

Ylarod:
Not tested, let me try it first

Ylarod:
```
rubens:/ # find /sys -name "ksu_*"
/sys/module/apk_sign/parameters/ksu_expected_hash
/sys/module/apk_sign/parameters/ksu_expected_size
```

tiann (approved):

Ylarod:
```
rubens:/ # find /sys -name "ksu_*"
/sys/module/kernelsu/parameters/ksu_expected_hash
/sys/module/kernelsu/parameters/ksu_expected_size
```

tiann:
looks great!

PR #25, opened 2023-01-05
Kconfig : Make KSU depends on OVERLAY_FS

hmtheboy154:
As KernelSU now support modules and mounting system as R/W using overlayfs, it's time to require overlayfs when trying to compile it.


tiann:
Thank you!

PR #37, opened 2023-01-10
kernel: remove warning

Ylarod:


kernel/allowlist.h
```diff
@@ -1,6 +1,8 @@
 #ifndef __KSU_H_ALLOWLIST
 #define __KSU_H_ALLOWLIST
 
+#include "linux/types.h"
```
tiann, line 4: please use `#include <linux/types.h>`

Ylarod: done

tiann (approved):

PR #46, opened 2023-01-12
kernel: We don't like dirty

lateautumn233:


tiann:
Can we put the patch files to `.github` directory? It may confuse the developer if they check the source code.

Scirese:
put them into /patches and add a suitable Readme.

tiann:
this is a CI specific, put it to .github is better.

tiann (approved):

PR #48, opened 2023-01-12
kernel: Change CI build for WSA-5.10.117

longhuan1999:
1. The WSA kernel version has nothing to do with the Android version. Remove the Android version display
2. Merge x86_ 64 and arm64 strategy

### DIFF
--- a/.github/workflows/build-WSA-5.10.117-kernel.yml
+++ b/.github/workflows/build-WSA-5.10.117-kernel.yml
@@ -64,6 +64,8 @@ jobs:
         echo "[+] 添加 kernel su driver 到文件：$DRIVER_MAKEFILE"
         grep -q "kernelsu" $DRIVER_MAKEFILE || echo "obj-y += kernelsu/" >> $DRIVER_MAKEFILE
         echo "[+] KernelSU 导入完成"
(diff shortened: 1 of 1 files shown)


tiann:
By the way, can we just build the latest kernel?

longhuan1999:
Before WSA use kernel 5.10.117.4-20221110(Android12), but now WSA use 5.10.117.2-20220906(Android13). I'm not sure which one they will choice in the next time.

longhuan1999:
However, for the time being, we can choice  just build 5.10.117.2-20220906

tiann:
Can any of a WSA kernel boot successfully?

tiann:
What if the kernel version doesn't match? boot failed?

longhuan1999:
I haven't done such tests.

tiann:
Can you modify this PR again? After this PR, the compile time is twice as before, because we build arm64 and x86_64 in a job, but it was two job before. Although we half the `git clone`, we twice the compile time

longhuan1999:
In this case, then how about modifing it to only compile the kernel version currently used by WAS at the same time?

tiann:
i think this is a good idea

PR #49, opened 2023-01-13
kernel: add prefix for functions in sepolicy

Ylarod:


tiann:
can we use compiler extension like __always_inline？

Ylarod:
ofc, but I think `static inline` is enough to solve problems like this
```
drivers/kernelsu/selinux/sepolicy.o: In function `allow':
sepolicy.c:(.text+0xec0): multiple definition of `allow'
drivers/clk/mediatek/clk-mt6853-pg.o:(.data+0x10): first defined here
/home/chen/android_kernel_xiaomi_cannon/Makefile:1144: recipe for target 'vmlinux' failed
make[1]: *** [vmlinux] Error 1
make[1]: Leaving directory '/home/chen/android_kernel_xiaomi_cannon/out'
Makefile:146: recipe for target 'sub-make' failed
make: *** [sub-make] Error 2
```

tiann:
using `__always_inline`, we don't need to put all things to header file, and it is easier to maintain(maybe one day we want to export it?)

Ylarod:
emm.. how to ?

tiann (approved):

tiann (approved):

PR #58, opened 2023-01-14
kernel: export customized symbols

Ylarod:


.github/workflows/build-kernel-5.10.yml
```diff
@@ -56,13 +66,22 @@ jobs:
...
+        echo "[+] KernelSU setup Done."
     
+    - name: Symbol magic
+      run: |
+        echo "[+] Export all symbol from abi_gki_aarch64.xml"
+        COMMON_ROOT=$GITHUB_WORKSPACE/android-kernel/common
+        KSU_ROOT=$GITHUB_WORKSPACE/KernelSU
+        ABI_XML=$COMMON_ROOT/android/abi_gki_aarch64.xml
+        SYMBOL_LIST=$COMMON_ROOT/android/abi_gki_aarch64
+        python3 $KSU_ROOT/scripts/abi_gki_all.py $ABI_XML > $SYMBOL_LIST
```
tiann, line 81: I think we shouldn't export all symbols now, if mod developers need this, we can add it in the future.

tiann (changes requested):

.github/workflows/build-kernel-5.10.yml
Ylarod: this export all symbols in abi_gki_aarch64.xml, but this xml doesn't contains all exported symbols

tiann (approved):

Ylarod:
`$SYMBOL_LIST` only contains two symbols.
```
[abi_symbol_list]
# commonly used symbols
  module_layout
  __put_task_struct
```
Just append our symbol to it is ok
Regenerate it is also acceptable.

PR #63, opened 2023-01-14
kernel: allow root processes to get version

Scirese:


kernel/ksu.c
```diff
@@ -244,6 +244,16 @@ static int handler_pre(struct kprobe *p, struct pt_regs *regs)
 		return 0;
 	}
 
+	// Both root manager and root processes should be allowed to get version
+	if( __manager_uid == current_uid().val || 0 == current_uid().val ) {
```
tiann, line 248: Check the arg2 first, and then check if it is valid role.
And also, please reuse the `is_manager`, don't repeat yourself.

kernel/ksu.c
```diff
@@ -244,6 +244,16 @@ static int handler_pre(struct kprobe *p, struct pt_regs *regs)
...
+		if (arg2 == CMD_GET_VERSION) {
+			u32 version = KERNEL_SU_VERSION;
+			if (copy_to_user(arg3, &version, sizeof(version))) {
+				pr_err("prctl reply error, cmd: %d\n", arg2);
+			}
```
tiann, line 253: we should return here.

tiann (changes requested):

Scirese:
```	
// Both root manager and root processes should be allowed to get version
	if (arg2 == CMD_GET_VERSION) {
		if( is_manager() || 0 == current_uid().val ) {
			u32 version = KERNEL_SU_VERSION;
			if (copy_to_user(arg3, &version, sizeof(version))) {
				pr_err("prctl reply error, cmd: %d\n", arg2);
				return 0;
			}
		}
	}
```
Like this?

tiann:
> ```
> // Both root manager and root processes should be allowed to get version
> 	if (arg2 == CMD_GET_VERSION) {
> 		if( is_manager() || 0 == current_uid().val ) {
> 			u32 version = KERNEL_SU_VERSION;
> 			if (copy_to_user(arg3, &version, sizeof(version))) {
> 				pr_err("prctl reply error, cmd: %d\n", arg2);
> 				return 0;
> 			}
> 		}
> 	}
> ```
> 
> Like this? ~(let me sure before pushing commits)~

looks good

tiann:
But the second if should be formatted.

Scirese:
OK

kernel/ksu.c
```diff
@@ -244,6 +244,17 @@ static int handler_pre(struct kprobe *p, struct pt_regs *regs)
 		return 0;
 	}
 
+	// Both root manager and root processes should be allowed to get version
+	if (arg2 == CMD_GET_VERSION) {
+		if( is_manager() || 0 == current_uid().val ) {
+			u32 version = KERNEL_SU_VERSION;
+				if (copy_to_user(arg3, &version, sizeof(version))) {
```
Ylarod, line 251: Incorrect indent

Ylarod (changes requested):

kernel/ksu.c
Scirese: I don't want CI to rerun, could you please fix it after merge?

tiann: It is ok to let it run

Ylarod: add `[skip ci]` to your commit message
https://docs.github.com/zh/actions/managing-workflow-runs/skipping-workflow-runs

Ylarod (approved):

tiann (approved):

PR #66, opened 2023-01-14
Use getuid() to check elevation status

Scirese:
Seems ksuctl()'s feedback is unreliable, we still need getuid!

 Sorry for not realizing this problem even someone told me🥲

### DIFF
--- a/userspace/su/jni/su.c
+++ b/userspace/su/jni/su.c
@@ -21,20 +21,20 @@ static void help(int status) {
 
 }
 
-static bool ksuctl(int cmd, void * arg1, void * arg2) {
+void ksuctl(int cmd, void * arg1, void * arg2) {
 
     int32_t result = 0;
     prctl(0xDEADBEEF, cmd, arg1, arg2, & result);
-    return result == 0xDEADBEEF;
+
 
 }
 
 void elevate() {
 
     // Talk to Daemon in Kernel Space
-    bool status = ksuctl(CMD_GRANT_ROOT, 0, NULL);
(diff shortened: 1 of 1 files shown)


tiann:
Why it is unreliable?

Scirese:
> Why it is unreliable?

A manual breakpoint shows that it will always return `false`. Regardless of the elevation is successful or not.

tiann:
We should make it reliable instead of bypass it

Scirese:
> We should make it reliable instead of bypass it

I need to study it and that will take some time, what about use this as a temporary solution?

tiann:
I will check it later

tiann:
temporary = won't fix

Scirese:
A Hex Integer `0xdeadbeef` should be `-559038737` in decimal, but prctl's `result` seems don't give a fixed value for successful or not🤔, so the problem might in how kernel replies prctl?

tiann:
What is the value of result ?

Scirese:
```
static bool ksuctl(int cmd, void* arg1, void* arg2) {

    int32_t result = 0;
    prctl(0xDEADBEEF, cmd, arg1, arg2, &result);
    return result == 0xDEADBEEF;

}
```

Variable `result` in this function

Ylarod:
No reply here, add it in kernel
https://github.com/tiann/KernelSU/blob/main/kernel/ksu.c#L235

Scirese:
Seems issue isn't here. Closing this PR.

PR #74, opened 2023-01-16
manager: Update su

Scirese:


Ylarod:
Don't repeat yourself.
Build userspace/su and move it to manager.

Scirese:
> Don't repeat yourself. Build userspace/su and move it to manager.

Is this allowed by manager？

tiann:
manager's su don't need this

PR #79, opened 2023-01-17
kernel: fix reboot caused by root_groups

isBaibai:
After using the latest version of kernelSU, it often reboots, the way to reproduce: after using any application that requires Runtime.getRuntime().exec("su"), close it, reopen it and try to get permission.
After several tests, I found that the problem occurs after modifying init_groups, and after rolling back the code, the compiled kernel runs fine.
Device Redmi/diting, kernel version 5.10.101

### DIFF
--- a/kernel/ksu.c
+++ b/kernel/ksu.c
@@ -30,7 +30,9 @@
 #include "selinux/selinux.h"
 #include "uid_observer.h"
 
-static struct group_info root_groups = { .usage = ATOMIC_INIT(2) };
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 15, 0)
+	static struct group_info root_groups = { .usage = ATOMIC_INIT(2) };
+#endif
 
 uid_t ksu_manager_uid = INVALID_UID;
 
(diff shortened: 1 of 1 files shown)


kernel/ksu.c
```diff
@@ -30,7 +30,9 @@
 #include "selinux/selinux.h"
 #include "uid_observer.h"
 
-static struct group_info root_groups = { .usage = ATOMIC_INIT(2) };
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 15, 0)
+	static struct group_info root_groups = { .usage = ATOMIC_INIT(2) };
```
Ylarod, line 34: no indent here

Ylarod (changes requested):
One code format problem.
We need futher information to investigate this issue.

natsumerinchan:
[pstore.zip](https://github.com/tiann/KernelSU/files/10435379/pstore.zip)

Oneplus 9RT has the same problem.

Ylarod:
fix in #80

PR #87, opened 2023-01-19
manager: Get SELinux status and display it at home

xiaoleGun:
Sorry, there may be a better way to do this, but I didn't think of it.

Tested fine on vivo, huawei and lineageos.

### DIFF
--- a/manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
+++ b/manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
@@ -37,6 +37,7 @@ import me.weishu.kernelsu.ui.screen.destinations.SettingScreenDestination
 import me.weishu.kernelsu.ui.util.LinkifyText
 import me.weishu.kernelsu.ui.util.LocalSnackbarHost
 import me.weishu.kernelsu.ui.util.reboot
+import me.weishu.kernelsu.ui.util.SELinuxStatus
 
 @OptIn(ExperimentalMaterial3Api::class)
 @RootNavGraph(start = true)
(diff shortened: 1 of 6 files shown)


manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
```diff
@@ -200,6 +200,26 @@ private fun InfoCard() {
             val contents = StringBuilder()
             val uname = Os.uname()
 
+            val GetSELinuxStatus = Runtime.getRuntime().exec("getenforce")
```
tiann, line 203: Please correct variable name

tiann: Please follow the java naming convention, do not start with an uppercase letter

tiann: Put this to a single method, because we may change it implementation in the future(use libselinux instead)

tiann (changes requested):

manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
xiaoleGun: ok, I will do

manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
```diff
@@ -200,6 +200,26 @@ private fun InfoCard() {
...
+            GetSELinuxStatus.waitFor()
+            val GetSELinuxStatusOut = GetSELinuxStatus.errorStream.bufferedReader().readLine() ?: GetSELinuxStatus.inputStream.bufferedReader().readLine()
```
Dr-TSNG, line 205: stream not closed?

Dr-TSNG (changes requested):

manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
xiaoleGun: thank you for correction i will close it after getting the results

manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
xiaoleGun: has made it a method and change variable to start with a lowercase letter

tiann:
How about use [libsu](https://github.com/topjohnwu/libsu)? We already has dependency on it and don't need to create process manually.

PR #90, opened 2023-01-20
modify to support non-gki

logic10492:


Ylarod:
Non-GKI devices won‘t be officially supported

PR #98, opened 2023-01-21
Add SELinux status

tiann:


tiann:
close #87

xiaoleGun:
I want to change to libsu library after Chinese New Year's eve dinner :)

PR #103, opened 2023-01-22
README: Added the information about the supoorting of Linux 4.9

SlightlyLookAround:
- And Linux 4.14 is same. See the disscusion of https://www.coolapk.com/feed/42545715.


tiann:
why and how we should modify kernel source？

tiann:
why and how we should modify kernel source？

Ylarod:
that patch in qq channel?

tiann:
4.14 doesn't have selinux_state, it shouldn't work now

SlightlyLookAround:
> why and how we should modify kernel source？

The references of selinux_state(). We found out the backport on Linux 4.14 and 4.9.
For example, from https://github.com/SlightlyLookAround/kernel_for_vince/commit/3282dcf1ffaba204f1933dc7a695e2269cebf707 to https://github.com/SlightlyLookAround/kernel_for_vince/commit/f3051234e89ddd1666f15206356f54c8bfe5968c. And they works very well.

tiann:
modify KernelSU to make it work on old kernel ✅
modify kernel to make it work on KernelSU ❌

SlightlyLookAround:
> modify KernelSU to make it work on old kernel ✅
> modify kernel to make it work on KernelSU ❌

Ikr, but how truly is that this is another solution. 😂

Scirese:
I tried to do some backport on 4.9. Seems most kernel part could work(including selinux), but it seems to have some problem on recgonizing manager, here's some interesting log:
```
[    4.741239] (6)[1:swapper/0]KernelSU: newfstatat_kp: -2
[   27.920608] (1)[74:kworker/1:1]KernelSU: do_update_uid, open /data/system/packages.list failed: -126
[   62.888157] (2)[337:kworker/2:1]KernelSU: save_allow_list creat file failed: -126
[   63.547150] (2)[6785:weishu.kernelsu]KernelSU: error: cannot find eocd
[   63.547158] (5)[1:init]KernelSU: manager signature invalid!
```
Full log: [2.txt](https://github.com/tiann/KernelSU/files/10473777/2.txt)
I think the problem is on `kernel_read` and `kernel_write` call, on 4.9 and 4.4 they have a diffrent argument type so KernelSU couldn't write allowlist and read manager.

On my personal perspective, modifying kernel source seems unavoidable.

xiaoleGun:
Well, `selinux_state` backported on android-4.14-common(ACK), so, most 4.14 kernel already have `selinux_state`

We should consider how to make kernelsu work without `selinux_state` in 4.9

tiann:
我们会让走不动路的老奶奶使用轮椅，而不是给她换一双腿。旧内核有旧内核的工作方式，以它的方式工作就行了。

tiann:
我们拿 selinux state 是为了访问 policydb，难道旧内核没有 policydb？

Ylarod:
revert this changes ?

tiann:
如果将来 Linux 6.5 把 selinux_state 改个名或者放到另外一个结构体，你是不是要把 5.x 的内核代码 pick 到 6.x？

Scirese:
On 4.9 policydb existed standalone,
We only need:
```
db = &policydb;
```
and it will work.

tiann:
> revert this change ?

better do that

SlightlyLookAround:
I haven't idea about this disscusion, just did what I can do, so go for it and wish you happy for the new year.😂
K thx bye👋🏻

tiann:
we will make it work on 4.9 in the future, but i don't think this is the correct way

ookiineko:
i have also tried 4.9 and it worked on my device [without kernel side changes](https://github.com/dandelion64-Archives/android_kernel_xiaomi_dandelion/tree/staging/ksu)



but with one ugly hack for now:

image

image


patches: https://github.com/ookiineko/KernelSU/commits/main

i have only tested adb shell and termux



image

ookiineko:
image

if your 4.9 kernel already has statx syscall backported or sth, just ignore this patch i guess

ookiineko:
> i have also tried 4.9 and it worked on my device

i forgot to say, there's still some bugs obviously:

1. allowlist isn't saved on reboot (bcs reading and writing both failed for some reason
2. signature verification for manager app is broken (that is why i used a hack, i don't know what is wrong with it

ookiineko:
> 1. allowlist isn't saved on reboot (bcs reading and writing both failed for some reason

although an empty file was created

image

ookiineko:
bruh, kernel_read changed a lot and i didn't realize it, probably this is the reason

tiann:
> bruh, kernel_read changed a lot and i didn't realize it, probably this is the reason

```c
static ssize_t kernel_read_compat(struct file *p, void* buf, size_t count, loff_t *pos) {
#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 14, 0)
    return kernel_read(p, buf, count, pos);
#else
    loff_t offset = pos ? *pos : 0;
    ssize_t result = kernel_read(p, offset, (char *) buf, count);
    if (pos && result > 0) {
        *pos = offset + result;
    }
    return result;
#endif
}

static ssize_t kernel_write_compat(struct file *p, const void *buf, size_t count, loff_t *pos) {
#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 14, 0)
    return kernel_write(p, buf, count, pos);
#else
    loff_t offset = pos ? *pos : 0;
    ssize_t result = kernel_write(p, buf, count, offset);
    if (pos && result > 0) {
        *pos = offset + result;
    }
    return result;
#endif
}
```

Add this and replace original `kernel_read/write` to `kernel_read/write_compat`

ookiineko:
> Add this and replace original `kernel_read/write` to `kernel_read/write_compat`

ok, let me try it

ookiineko:
oh, i so dereferenced NULL pointer sometimes?

and the offset is incremented in the later version of kernel_read/write?

ookiineko:
ohhh, the weird bug is gone and now my device boots normally (it was not booting when i first try to fix the arg order of kernel_read, but with your advice i am able to get it booting

ookiineko:
yesss, no need to apply the ugly hack anymore

ookiineko:
the kernel log is very clean now

image

PR #104, opened 2023-01-23
manager: Add monochrome themed icon on Android 12+

AngelaCooljx:
The icon looks more modern when using themed icon on Android 12+


tiann:
Thank you!

PR #106, opened 2023-01-23
website:Remind readers how to resolve dependencies for KPROBES

natsumerinchan:
MODULES is required for KPROBE but some kernel disabled it.

### DIFF
--- a/kernel/Kconfig
+++ b/kernel/Kconfig
@@ -1,7 +1,6 @@
 config KSU
 	tristate "KernelSU module"
 	default y
-	depends on MODULES
 	depends on KPROBES
 	depends on OVERLAY_FS
 	help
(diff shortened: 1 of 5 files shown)


Ylarod:
Kernel won't resolve dependencies automatically ?
KRETPROBES depends on KPROBES, but it doesn't depend on MODULES.

kernel/Kconfig
```diff
@@ -1,6 +1,7 @@
 config KSU
 	tristate "KernelSU module"
 	default y
+	depends on MODULES
```
Ylarod, line 4: I don't think this is needed

Ylarod (changes requested):

kernel/Kconfig
natsumerinchan: You said makes sense, I'll revert these commits.Maybe we just need to tell readers about the KPROBES dependencies on the website and let them debug by themselves.

PR #120, opened 2023-01-26
ci: use custom key to sign official bootimgs

Ylarod:
We can get pub key hash by 
```
avbtool info_image --image boot.img
```


tiann:
how to generate `BOOT_SIGN_KEY` ?

Ylarod:
```
openssl genrsa -out rsa_private_key.pem 2048
```

tiann:
Is Magisk use a custom key to sign boot or only use test key?

Ylarod:
> Is Magisk use a custom key to sign boot or only use test key?

seems no.
https://github.com/topjohnwu/Magisk/blob/master/tools/keys/

PR #123, opened 2023-01-26
some Kconfig patches

ookiineko:


tiann:
how the config is used？

ookiineko:
in `menuconfig`

ookiineko:
CONFIG_OVERLAY_FS is now automatically selected by CONFIG_KSU

ookiineko:
when CONFIG_KSU_NO_KPROBE is enabled, users can leave CONFIG_KPROBES disabled

ookiineko:
i also changed the option type of CONFIG_KSU_DEBUG from tristate to bool, bcs it is just a debug flag, should never be set to =m

ookiineko:
bcs it does not make sense

ookiineko:
in that case

ookiineko:
> when CONFIG_KSU_NO_KPROBE is enabled, users can leave CONFIG_KPROBES disabled

without CONFIG_KSU_NO_KPROBE, users won't be able to find CONFIG_KSU in the kernel config menu unless they turn the CONFIG_KPROBES on

ookiineko:
demo

[2023-01-27 13-02-47.mkv.tar.gz](https://github.com/tiann/KernelSU/files/10515749/2023-01-27.13-02-47.mkv.tar.gz)

ookiineko:
changed the type of CONFIG_KSU to bool as well, or need changes to the Makefiles

but is it actually useful if KSU is compiled as a kernel module?

tiann:
Just modify the Kconfig is not enough, because this config is unused in main branch.

ookiineko:
> Just modify the Kconfig is not enough, because this config is unused in main branch.

true, need to modify the Makefile as well, closing for now

PR #124, opened 2023-01-26
kernel: Makefile: add -Wno-error

ookiineko:
some retarded (SoC) vendor adds -Werror to their kernel root Makefile :(


ookiineko:
this is due to the https://github.com/tiann/KernelSU/blob/6e8771b8d939cbffa20ad4923c416de7b1a6be61/kernel/ksu.c#L60

hmtheboy154:
wow that is sad........ but add it into the Documentation instead of pushing this

ookiineko:
ok tku

PR #128, opened 2023-01-28
docs: add unofficially-support-devices

Ylarod:


tiann:
maybe we should also list the device name or something to identify the device

tiann:
we should also add disclaimer, the kernel's issue has nothing to do with KernelSU.

tiann (approved):

PR #154, opened 2023-01-30
Website: add Asus Zenfone Max Pro M1 (X00TD)

ajugeorge98:


Ylarod (changes requested):
edit website/docs/repos.json now

PR #155, opened 2023-01-30
website: fix a syntax error

YamazakuraH:


tiann:
The format changed, this doesn't need now.

PR #160, opened 2023-01-31
kernel: allowlist v2

Ylarod:


kernel/allowlist.c
```diff
@@ -1,25 +1,40 @@
+#include "asm-generic/int-ll64.h"
```
tiann, line 1: It this we need?

kernel/allowlist.c
```diff
@@ -1,25 +1,40 @@
...
 #include "selinux/selinux.h"
 
 #define FILE_MAGIC 0x7f4b5355 // ' KSU', u32
-#define FILE_FORMAT_VERSION 1 // u32
+#define FILE_FORMAT_VERSION 2 // u32
 
 static DEFINE_MUTEX(allowlist_mutex);
 
-struct perm_data {
+const struct perm_data NO_PERM = {
```
tiann, line 20: maybe static?

kernel/allowlist.c
```diff
@@ -1,25 +1,40 @@
...
+	.allow = false,
+	.cap = CAP_EMPTY_SET,
+};
+
+const struct perm_data ALL_PERM = {
```
tiann, line 25: static

kernel/allowlist.c
```diff
@@ -28,42 +43,56 @@ static struct work_struct ksu_load_work;
 
 bool persistent_allow_list(void);
 
+void ksu_show_perm_data(struct perm_data data)
```
tiann, line 46: use `struct perm_data *`

kernel/allowlist.c
```diff
@@ -28,42 +43,56 @@ static struct work_struct ksu_load_work;
...
+{
+	pr_cont("allow: %d, cap: %08x | %08x", data.allow, data.cap.cap[0],
+		data.cap.cap[1]);
+}
+
+void ksu_show_perm_list_node(const char* str, struct perm_list_node data)
```
tiann, line 52: use pointer

kernel/allowlist.c
tiann: and static

kernel/allowlist.c
tiann: static

tiann: rename it to node to avoid `data->data`

kernel/allowlist.c
```diff
@@ -73,49 +102,74 @@ bool ksu_allow_uid(uid_t uid, bool allow, bool persist)
 	return result;
 }
 
-bool ksu_is_allow_uid(uid_t uid)
+struct perm_data ksu_get_uid_data(uid_t uid)
```
tiann, line 105: `ksu_get_uid_perm` is better, so many "data" here, it is confusing

tiann (changes requested):

kernel/allowlist.c
Ylarod: exposed to other part, cannot be static

PR #162, opened 2023-01-31
ci: replace actions-rs/cargo with naive command

skbeh:
This avoids node 12 warnings.

### DIFF
--- a/.github/workflows/ksud.yml
+++ b/.github/workflows/ksud.yml
@@ -29,11 +29,7 @@ jobs:
         restore-keys: ${{ runner.os }}-cargo-
 
     - name: Build ksud
-      uses: actions-rs/cargo@v1
-      with:
-        use-cross: true
-        command: build
-        args: --target ${{ inputs.target }} --release --manifest-path ./userspace/ksud/Cargo.toml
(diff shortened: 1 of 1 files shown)


tiann (approved):

tiann:
This doesn't trigger CI?

.github/workflows/ksud.yml
```diff
@@ -29,11 +29,7 @@ jobs:
         restore-keys: ${{ runner.os }}-cargo-
 
     - name: Build ksud
-      uses: actions-rs/cargo@v1
-      with:
-        use-cross: true
-        command: build
-        args: --target ${{ inputs.target }} --release --manifest-path ./userspace/ksud/Cargo.toml
+      run: cross build --target ${{ inputs.target }} --release --manifest-path ./userspace/ksud/Cargo.toml
```
tiann, line 32: when the `cross` is installed?

Ylarod: seems built-in in action's ubuntu image

skbeh: According to https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md, it is included in the action image.

PR #167, opened 2023-02-01
ksud: fix issues found by `clippy`

skbeh:
These issues are mostly found by `cargo clippy -- -W clippy::pedantic`.

### DIFF
--- a/userspace/ksud/src/event.rs
+++ b/userspace/ksud/src/event.rs
@@ -1,11 +1,11 @@
+use anyhow::{bail, Context, Result};
+use log::{info, warn};
 use std::{collections::HashMap, path::Path};
 
 use crate::{
     assets, defs, mount,
     utils::{ensure_clean_dir, ensure_dir_exists},
 };
-use anyhow::{bail, Context, Result};
-use log::{info, warn};
 
 fn mount_partition(partition: &str, lowerdir: &mut Vec<String>) -> Result<()> {
     if lowerdir.is_empty() {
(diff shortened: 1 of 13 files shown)


userspace/ksud/src/event.rs
```diff
@@ -149,16 +151,14 @@ pub fn on_post_data_fs() -> Result<()> {
...
-    } else {
-        warn!("safe mode, skip module post-fs-data scripts");
     }
 
     Ok(())
 }
 
 pub fn on_services() -> Result<()> {
     // exec modules service.sh scripts
-    if !crate::utils::is_safe_mode() {
```
tiann, line 161: this is wrong

userspace/ksud/src/module.rs
```diff
@@ -197,6 +189,40 @@ pub fn load_sepolicy_rule() -> Result<()> {
...
+    command = command
+        .current_dir(path.as_ref().parent().unwrap())
+        .arg("sh")
+        .arg(path.as_ref())
+        .env("ASH_STANDALONE", "1")
+        .env(
+            "PATH",
+            format!("{}:{}", env_var("PATH").unwrap(), defs::BINARY_DIR),
+        )
+        .env("KSU", "1");
```
tiann, line 216: should be "KSU" "true" instead "1"

tiann (changes requested):

userspace/ksud/src/module.rs
tiann: please modify this, some modules already depend on this behavior

tiann (approved):

skbeh:
@tiann Is the [cast to `i8`](https://github.com/tiann/KernelSU/blob/219ea1c458fb356b12fc471c1470ffcd2bad29df/userspace/ksud/src/apk_sign.rs#L85) and the [cast to `u32`](https://github.com/tiann/KernelSU/blob/219ea1c458fb356b12fc471c1470ffcd2bad29df/userspace/ksud/src/apk_sign.rs#L91) supposed to overflow?

tiann:
> cast to `i8`

Yes, it is expected behavior

PR #166, opened 2023-02-01
kernel: changes for old Kernel

F-19-F:
These changes make KernelSU work on kernel4.4 [link](https://github.com/F-19-F/android_kernel_oneplus_msm8998).
LINUX_VERSION_CODE macro changes have been vertied on 4.4 4.9 4.14.
For kernel 4.4,just pick two commits
* [introduce KernelSU](https://github.com/F-19-F/android_kernel_oneplus_msm8998/commit/2993524f2fe7fbd2eea28883ec9f36e86de50f85)
* [allow init exec ksud under nosuid](https://github.com/F-19-F/android_kernel_oneplus_msm8998/commit/3df9df42a659f489091445d813b8c0477215ae3a)

PR #174, opened 2023-02-01
kernel: more changes for old kernels

ookiineko:
this adds some missing bits and also corrects sth in the previous PR #166


ookiineko:
oops

PR #186, opened 2023-02-03
ksud: compress embed assets

skbeh:


userspace/ksud/src/assets.rs
```diff
@@ -1,22 +1,28 @@
...
+struct Asset;
 
-pub fn ensure_bin_assets() -> Result<()> {
-    utils::ensure_binary(RESETPROP_PATH, RESETPROP)?;
-    utils::ensure_binary(BUSYBOX_PATH, BUSYBOX)?;
+pub fn ensure_binaries() -> Result<()> {
+    for file in Asset::iter() {
+        utils::ensure_binary(
+            format!("{BINARY_DIR}{file}"),
+            &Asset::get(&file).unwrap().data,
```
tiann, line 24: can we use `?` instead of "unwrap"?

tiann (changes requested):

userspace/ksud/src/assets.rs
skbeh: The file certainly exists because it is provided from the embedded file list.

tiann (approved):

PR #194, opened 2023-02-03
ksud, kernel: update explanation and fix some issues

skbeh:


userspace/ksud/src/installer.sh
```diff
@@ -1,4 +1,5 @@
 #!/system/bin/sh
+# shellcheck disable=SC3043
```
tiann, line 2: don't touch this file, most of this is from Magisk, it is stable enough.

tiann (changes requested):

tiann (approved):

PR #211, opened 2023-02-06
kernel: get remote commits

Ylarod:


tiann:
1. If .git not found, fetch it from remote.
2. just put the script in kernel folder, some guys just copy this directory to his kernel.

tiann:
https://gist.github.com/0penBrain/7be59a48aba778c955d992aa69e524c5  this script don't need to use script.

PR #215, opened 2023-02-07
kernel: selinux: fix version check for ACK and qcom kernels

ookiineko:
in 3.14, qcom added a change that removes "#include "avc.h"" from objsec.h,
this change was dropped since 4.14.
    
4.14 ack has selinux_state backported by Google (which we should mainly target).

### DIFF
--- a/kernel/selinux/selinux.c
+++ b/kernel/selinux/selinux.c
@@ -2,7 +2,8 @@
 #include "objsec.h"
 #include "linux/version.h"
 #include "../klog.h" // IWYU pragma: keep
-#if LINUX_VERSION_CODE < KERNEL_VERSION(4, 17, 0)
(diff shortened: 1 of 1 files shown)


ookiineko:
this fixes #181

tiann (approved):

tiann:
https://github.com/tiann/KernelSU/pull/214 this is better

ookiineko:
what about the first part?

> in 3.14, qcom added a change that removes "#include "avc.h"" from objsec.h,
> this change was dropped since 4.14.

i think we can remove the version check as well since avc.h has include guard

tiann:
> what about the first part?
> 
> > in 3.14, qcom added a change that removes "#include "avc.h"" from objsec.h,
> > this change was dropped since 4.14.
> 
> i think we can remove the version check as well since avc.h has include guard

It is good

PR #230, opened 2023-02-11
ksud: remove tailing "/" from `PATH`

skbeh:


tiann:
tailing with `/` doesn't work?

skbeh:
It works, but it isn't the common practice.

tiann:
I someone edit `BINARY_DIR` someday, error will happen. There isn't any mechanism to ensure this is correct. But if you leave it as it is, it would always work.

userspace/ksud/src/module.rs
```diff
@@ -44,7 +44,11 @@ fn exec_install_script(module_file: &str) -> Result<()> {
         .env("ASH_STANDALONE", "1")
         .env(
             "PATH",
-            format!("{}:{}", env_var("PATH").unwrap(), defs::BINARY_DIR),
+            format!(
+                "{}:{}",
+                env_var("PATH").unwrap(),
+                &defs::BINARY_DIR[..(defs::BINARY_DIR.len() - 1)] // remove tailing "/"
```
tiann, line 50: `trim_end_matches` would be better

tiann (changes requested):

tiann (approved):

PR #231, opened 2023-02-11
ksud: assume all scripts are executable

skbeh:
Interpreting bash scripts does not need executable permission.


tiann:
this is intended behavior: https://topjohnwu.github.io/Magisk/guides.html#boot-scripts

> Only executed if the script is set as executable (chmod +x script.sh)

PR #241, opened 2023-02-14
Update build-kernel-a12.yml

esureos:


.github/workflows/build-kernel-a12.yml
```diff
@@ -21,10 +21,10 @@ jobs:
     strategy:
       matrix:
         include:
-          - sub_level: 66
```
tiann, line 24: Don't delete old one

tiann: And append to the end

tiann (changes requested):

PR #255, opened 2023-02-19
manager: Add key to LazyColumn's items to prevent incorrect allow sta…

TinyHai:
在超级用户页面，当用户通过搜索结果修改了授权状态，退出搜索后显示的应用列表会出现授权状态显示不正确，需要刷新应用列表才能正确显示。


tiann:
some apps share the same uid, it is better to use package name to identity it

TinyHai:
> some apps share the same uid, it is better to use package name to identity it

You're right, I've replaced it with packageName

tiann (approved):

tiann:
Thanks for your contribution!

PR #266, opened 2023-02-21
setup: Soft link through relative path

easterNday:
In the process of using Github Actions, source code synchronization will generate some errors, and the absolute path method of soft links will cause the folders under the submodule to be unreadable, so this part has been updated to better adapt to Github Actions


tiann:
If you use git submodule, you should use `git submodule update` to pull source. This script is used for initializing the repo.
And i don't think use relative path is a good idea.

PR #291, opened 2023-03-03
manager: make a small adjustment to module screen

TinyHai:
The adjustment is as follows, the left is before the adjustment, the right is after the adjustment.

---
 
---


tiann:
I am sorry...I can't recognize the difference

TinyHai:
> I am sorry...I can't recognize the difference

It's really not easy to tell just by looking at the screenshots
In short, I moved the top and bottom padding into the LazyColumn to make the padding part scrollable

tiann (approved):

PR #294, opened 2023-03-04
ci: add secrets.APPLICATION_ID

longhuan1999:
Add secrets.APPLICATION_ ID to determine whether to customize the package name of the manager


tiann:
It is better to do something like this:https://github.com/tiann/KernelSU/blob/main/manager/sign.example.properties

modify the build.gradle and read APPLICATION_ID and NAME from properties, and edit the properties files from CI.

using sed is unstable.

PR #297, opened 2023-03-06
ci: add vars.APPLICATION_ID and vars.APPLICATION_NAME

longhuan1999:
Add vars.APPLICATION_ID and vars.APPLICATION_NAME to determine whether to customize the package name and app name of the manager


manager/build.gradle.kts
```diff
@@ -89,6 +89,24 @@ fun Project.configureBaseExtension() {
...
+            prop.load(signFile.inputStream())
+            var APPLICATION_ID = prop.getProperty("APPLICATION_ID")
+            var APPLICATION_NAME = prop.getProperty("APPLICATION_ID")
+            if (APPLICATION_ID.isBlank()){
+                APPLICATION_ID = "me.weishu.kernelsu"
+            }
+            if (APPLICATION_NAME.isBlank()){
+                APPLICATION_NAME = "KernelSU"
+            }
+            defaultConfig.applicationId = prop.getProperty("APPLICATION_ID")
```
tiann, line 103: it should be `defaultConfig.applicationId = APPLICATION_ID`

manager/build.gradle.kts
```diff
@@ -89,6 +89,24 @@ fun Project.configureBaseExtension() {
...
+            defaultConfig.resValue("string", "app_name", prop.getProperty("APPLICATION_NAME"))
```
tiann, line 104: should use `APPLICATION_NAME` instead.

manager/build.gradle.kts
```diff
@@ -89,6 +89,24 @@ fun Project.configureBaseExtension() {
             signingConfigs["debug"]
         }
 
+        if (signFile.canRead()) {
+            val prop = Properties()
+            prop.load(signFile.inputStream())
+            var APPLICATION_ID = prop.getProperty("APPLICATION_ID")
+            var APPLICATION_NAME = prop.getProperty("APPLICATION_ID")
+            if (APPLICATION_ID.isBlank()){
+                APPLICATION_ID = "me.weishu.kernelsu"
```
tiann, line 98: better to use a variable such as `DEFAULT_APPLICATION_ID`, because it is used later again.

tiann (changes requested):

manager/build.gradle.kts
longhuan1999: Sorry, I was careless

manager/build.gradle.kts
```diff
@@ -89,6 +89,27 @@ fun Project.configureBaseExtension() {
             signingConfigs["debug"]
         }
 
+        val DEFAULT_APPLICATION_ID = "me.weishu.kernelsu"
+        val DEFAULT_APPLICATION_NAME = "KernelSU"
+
+        if (signFile.canRead()) {
+            val prop = Properties()
+            prop.load(signFile.inputStream())
+            var APPLICATION_ID = prop.getProperty("APPLICATION_ID")
```
tiann, line 98: There is an overload method:

```
    public String getProperty(String key, String defaultValue) {
        String val = getProperty(key);
        return (val == null) ? defaultValue : val;
    }
```

So i think `prop.getProperty("APPLICATION_ID", DEFAULT_APPLICATION_ID)` is better.

tiann (changes requested):

PR #316, opened 2023-03-22
kernel: use git from PATH

shatyuka:
One should have git in PATH in order to build Android kernel.
Also it's weird that git is used in `setup.sh` without absolute path.


tiann:
When using bazel to build GKI kernel, the $PATH doesn't contain git

shatyuka:
so it's better to use /usr/bin/env?
something like this `/usr/bin/env PATH=$PATH:/usr/bin:/usr/local/bin git rev-list --count HEAD`

tiann:
> so it's better to use /usr/bin/env? something like this `/usr/bin/env PATH=$PATH:/usr/bin:/usr/local/bin git rev-list --count HEAD`

This should work

tiann:
I've tested it and it works.

PR #317, opened 2023-03-22
kernel: add new line in driver Makefile

shatyuka:
some device tree has no new line in Makefile


tiann (approved):

tiann:
reverted since there may be bugs: https://github.com/tiann/KernelSU/commit/6d6f793c6957abe1d9236a5a074a60c5319a1ee1

shatyuka:
got it, the posix `printf` should works for all distributions.

Issue #280, opened 2023-02-26
selinux_state not been back ported to mainline 4.9.212+


rhjdvsgsgks:
**Describe the bug**
#245 will use selinux_state after 4.9.212+ . but it only avaliable on android common kernel. and not on mainlane. it will cause issus when using mainline 4.14/4.9 kernel

**To Reproduce**
1. build ksu on top of https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable-rc.git/tag/?h=v4.14.274
2. get some error about selinux_state not found

**Expected behavior**
fallback to alternative method automatically

**Screenshots**

**Smartphone (please complete the following information):**
 - Device: any device using mainline kernel
 - OS: not a matter
 - Version: kernel 4.14.274

**Additional context**
works fine if i turn on the fallback method manually. i suggest add a way to control this manually by ccflags

tiann:
PRs welcome

tiann:
Due to the severe fragmentation of older kernel versions, any modifications made will not be able to satisfy everyone. However, if you have a better solution, welcome to make a PR : )

PR #324, opened 2023-03-26
selinux: Update 4.9 to the correct check

Dhruvgera:
4.9's last release aka 4.9.337 still needs the same fallbacks as 4.9.212 and breaks otherwise, upgrade this to fix compilation

Tested and working on POCO F1, 4.9.337


tiann:
https://github.com/tiann/KernelSU/issues/280

Dhruvgera:
> #280

As 4.9 is officially EOL and no longer upstreamed, the last revision is 4.9.337 and can be safely merged once as there will be no more future upgrades for this version in the future

PR #326, opened 2023-03-27
ci: Bump rustfmt-check to latest

Azure-Helper:


tiann:
I think it is better to use the latest release tag instead of  using master commit directly.

Azure-Helper:
> I think it is better to use the latest release tag instead of using master commit directly.

Done

tiann (approved):

PR #392, opened 2023-04-16
ksud: fix bind mount failed

5ec1cff:
This fixes https://github.com/Dr-TSNG/ZygiskOnKernelSU/issues/18


5ec1cff:
I found something wrong

PR #399, opened 2023-04-18
ci: Try to speed up kernel source checkout

Howard20181:


tiann (approved):

Howard20181:
Discontinued: The compilation toolchain is also part of KMI, and switching to the mirror on Github was unable to keep it the same.

PR #401, opened 2023-04-19
selinux: clamp 4.9 SUBLEVEL to 255

### LINKED ISSUE #400: Use 255 instead of 337
In #324, the version check starts using `KERNEL_VERSION(4, 9, 337)`, and from the reverts after the PR merge, it seems to break many 4.9 devices, and it did break my 4.9.337 Pixel 3.

My 4.9.337 Pixel 3 `linux/version.h`:
```cpp
#define LINUX_VERSION_CODE 264703
#define KERNEL_VERSION(a,b,c) (((a) << 16) + ((b) << 8) + (c))
```

The subversion number is only 8 bits, so it should only go up to 255. For some reason `LINUX_VERSION_CODE` follows the rule (`264703 == (4 << 16) + (9 << 8) + 255`) while `KERNEL_VERSION` doesn't. The newer `KERNEL_VERSION` is defined as follows, it seems some 4.9 devices haven't (and probably will never) been updated.
```cpp
#define KERNEL_VERSION(a,b,c) (((a) << 16) + ((b) << 8) + ((c) > 255 ? 255 : (c)))
```

On these 4.9 devices, the relevant conditions are never met, and to adapt, we can safely make changes like:
```diff
- #if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 9, 337)
+ #if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 9, 255)
- #if (LINUX_VERSION_CODE >= KERNEL_VERSION(4, 14, 163)) || ((KERNEL_VERSION(4, 10, 0) > LINUX_VERSION_CODE) && (LINUX_VERSION_CODE  >= KERNEL_VERSION(4, 9, 337)))
+ #if (LINUX_VERSION_CODE >= KERNEL_VERSION(4, 14, 163)) || (LINUX_VERSION_CODE == KERNEL_VERSION(4, 9, 255))
```

PaperStrike:
#324 is making the checks too strict that it is not requiring 4.9.337 but an updated `KERNEL_VERSION`. The official [4.9.337 Makefile](https://github.com/aosp-mirror/kernel_common/blob/f9b8314c64640cd10c7b14ce9d2a11a0dc02a941/Makefile#L1238) and even the latest [4.14](https://github.com/aosp-mirror/kernel_common/blob/3bfa20247ddad0861bd900b40e9c25e6f22c59dc/Makefile#L1337) and [4.19](https://github.com/aosp-mirror/kernel_common/blob/9b68392a2b5860995364db295d0d869b413fdfdf/Makefile#L1280) ones do not have a correct `KERNEL_VERSION` that clamps the subversion to 255 like their `LINUX_VERSION_CODE` do. The earliest kernel with a correct `KERNEL_VERSION` is 5.4 by [this commit](https://github.com/aosp-mirror/kernel_common/commit/e9be5518af2c1cd110824bd8b3327aedd9a80104). Comparing the 5.x commit with the 4.x ones (e.g. [this](https://github.com/aosp-mirror/kernel_common/commit/42efb098c75f9d4b95f7cf35d3d52e80cc2fde82)), it feels like an AOSP oversight but whatever we should adapt to it.

Fixes #400.

### DIFF
--- a/kernel/selinux/rules.c
+++ b/kernel/selinux/rules.c
@@ -22,7 +22,7 @@ static struct policydb *get_policydb(void)
 {
 	struct policydb *db;
 // selinux_state does not exists before 4.19
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 9, 337)
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 9, 255)
 #ifdef SELINUX_POLICY_INSTEAD_SELINUX_SS
 	struct selinux_policy *policy = rcu_dereference(selinux_state.policy);
 	db = &policy->policydb;
@@ -172,7 +172,7 @@ static void reset_avc_cache()
 {
 #if ((KERNEL_VERSION(4, 14, 0) <= LINUX_VERSION_CODE) &&                       \
      (LINUX_VERSION_CODE < KERNEL_VERSION(4, 14, 163))) ||                     \
-	(LINUX_VERSION_CODE < KERNEL_VERSION(4, 9, 337))
+	(LINUX_VERSION_CODE < KERNEL_VERSION(4, 9, 255))
 	avc_ss_reset(0);
 	selnl_notify_policyload(0);
 	selinux_status_update_policyload(0);
--- a/kernel/selinux/selinux.c
+++ b/kernel/selinux/selinux.c
@@ -2,7 +2,7 @@
 #include "objsec.h"
 #include "linux/version.h"
 #include "../klog.h" // IWYU pragma: keep
-#if ((KERNEL_VERSION(4, 14, 0) <= LINUX_VERSION_CODE) && (LINUX_VERSION_CODE < KERNEL_VERSION(4, 14, 163))) || (LINUX_VERSION_CODE < KERNEL_VERSION(4, 9, 337))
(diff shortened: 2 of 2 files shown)


PaperStrike:
It is not safe to merge yet and we had better to check the SUBLEVEL separately without relying on using `KERNEL_VERSION`.

PaperStrike:
@tiann Could we just test `selinux_state` in `kernel/selinux/Makefile`?

```makefile
ifeq ($(shell test -e $(srctree)/security/selinux/include/security.h; echo $$?),0)
ifeq ($(shell grep -q "extern struct selinux_state selinux_state;" $(srctree)/security/selinux/include/security.h; echo $$?),0)
ccflags-y += -DKSU_SELINUX_STATE
endif
endif
```

tiann:
> @tiann Could we just test `selinux_state` in `kernel/selinux/Makefile`?
> 
> ```makefile
> ifeq ($(shell test -e $(srctree)/security/selinux/include/security.h; echo $$?),0)
> ifeq ($(shell grep -q "struct selinux_state " $(srctree)/security/selinux/include/security.h; echo $$?),0)
> ccflags-y += -DKSU_SELINUX_STATE
> endif
> endif
> ```

I think this is a good idea, but please don't affect other kernel versions.

PR #402, opened 2023-04-19
selinux: use grep to check api supports

### LINKED ISSUE #280: selinux_state not been back ported to mainline 4.9.212+

### LINKED ISSUE #400: Use 255 instead of 337

PaperStrike:
This checks `selinux_state` and `current_sid` supports in a raw way. Feels more reliable than the version checks.

Supersedes #401, fixes #280, fixes #400.

### DIFF
--- a/kernel/selinux/Makefile
+++ b/kernel/selinux/Makefile
@@ -2,12 +2,13 @@ obj-y += selinux.o
 obj-y += sepolicy.o
 obj-y += rules.o
 
-ifeq ($(shell grep -q " current_sid(void)" $(srctree)/security/selinux/include/objsec.h; echo $$?),0)
-ccflags-y += -DKSU_COMPAT_HAS_CURRENT_SID
-endif
-
-ifeq ($(shell grep -q "struct selinux_state " $(srctree)/security/selinux/include/security.h; echo $$?),0)
-ccflags-y += -DKSU_COMPAT_USE_SELINUX_STATE
+ifeq ($(shell [ $(VERSION) -lt 5 -o \( $(VERSION) -eq 5 -a $(PATCHLEVEL) -lt 10 \) ]; echo $$?), 0)
+  ifeq ($(shell grep -q " current_sid(void)" $(srctree)/security/selinux/include/objsec.h; echo $$?),0)
+  ccflags-y += -DKSU_COMPAT_HAS_CURRENT_SID
+  endif
+  ifeq ($(shell grep -q "struct selinux_state " $(srctree)/security/selinux/include/security.h; echo $$?),0)
+  ccflags-y += -DKSU_COMPAT_USE_SELINUX_STATE
+  endif
 endif
 
 ccflags-y += -Wno-implicit-function-declaration -Wno-strict-prototypes -Wno-int-conversion
--- a/kernel/selinux/Makefile
+++ b/kernel/selinux/Makefile
@@ -2,7 +2,7 @@ obj-y += selinux.o
 obj-y += sepolicy.o
 obj-y += rules.o
 
-ifeq ($(shell [ $(VERSION) -lt 5 -o \( $(VERSION) -eq 5 -a $(PATCHLEVEL) -lt 10 \) ]; echo $$?), 0)
+ifeq ($(shell [ $(VERSION) -lt 5 -o \( $(VERSION) -eq 5 -a $(PATCHLEVEL) -lt 10 \) ]; echo $$?),0)
   ifeq ($(shell grep -q " current_sid(void)" $(srctree)/security/selinux/include/objsec.h; echo $$?),0)
   ccflags-y += -DKSU_COMPAT_HAS_CURRENT_SID
   endif
--- a/kernel/selinux/Makefile
+++ b/kernel/selinux/Makefile
@@ -7,7 +7,7 @@ ccflags-y += -DKSU_COMPAT_HAS_CURRENT_SID
 endif
 
 ifeq ($(shell grep -q "struct selinux_state " $(srctree)/security/selinux/include/security.h; echo $$?),0)
-ccflags-y += -DKSU_COMPAT_USE_SELINUX_STATE
+ccflags-y += -DKSU_COMPAT_HAS_SELINUX_STATE
 endif
 
 ccflags-y += -Wno-implicit-function-declaration -Wno-strict-prototypes -Wno-int-conversion
--- a/kernel/selinux/Makefile
+++ b/kernel/selinux/Makefile
@@ -2,6 +2,13 @@ obj-y += selinux.o
 obj-y += sepolicy.o
 obj-y += rules.o
 
+ifeq ($(shell grep -q " current_sid(void)" $(srctree)/security/selinux/include/objsec.h; echo $$?),0)
+ccflags-y += -DHAS_CURRENT_SID
+endif
+
+ifeq ($(shell grep -q "struct selinux_state " $(srctree)/security/selinux/include/security.h; echo $$?),0)
+ccflags-y += -DUSE_SELINUX_STATE
+endif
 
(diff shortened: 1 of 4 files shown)


kernel/selinux/Makefile
```diff
@@ -2,6 +2,13 @@ obj-y += selinux.o
 obj-y += sepolicy.o
 obj-y += rules.o
 
+ifeq ($(shell grep -q " current_sid(void)" $(srctree)/security/selinux/include/objsec.h; echo $$?),0)
+ccflags-y += -DHAS_CURRENT_SID
```
tiann, line 6: please prefix this MACRI with `KSU_` to avoid potential conflicts

kernel/selinux/rules.c
```diff
@@ -22,7 +22,7 @@ static struct policydb *get_policydb(void)
 {
 	struct policydb *db;
 // selinux_state does not exists before 4.19
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 9, 337)
+#ifdef USE_SELINUX_STATE
```
tiann, line 25: And also, don't rely on this MARCRO on GKI kernels (version bigger than 5.10)

tiann (changes requested):

kernel/selinux/Makefile
PaperStrike: https://github.com/tiann/KernelSU/pull/402/commits/47d21ee22a9f6b707246f2a06ec69814941d837f prefixed with `KSU_COMPAT_`

kernel/selinux/rules.c
PaperStrike: https://github.com/tiann/KernelSU/pull/402/commits/e4f43d1b605a1717f3993723497b81af34963508 ignored on 5.10+.
Maybe we can further optimize out the grep check in Makefile for 5.10+ by checking the `VERSION`/`PATCHLEVEL` which are set by the kernel root Makefile, but I am not quite familiar with Makefile syntax.

PaperStrike: https://github.com/tiann/KernelSU/pull/402/commits/36909422654b36db74e0574e3ead907697098cc8 fixed.

kernel/selinux/Makefile
```diff
@@ -2,6 +2,14 @@ obj-y += selinux.o
 obj-y += sepolicy.o
 obj-y += rules.o
 
+ifeq ($(shell [ $(VERSION) -lt 5 -o \( $(VERSION) -eq 5 -a $(PATCHLEVEL) -lt 10 \) ]; echo $$?),0)
```
tiann, line 5: I think this is unnecessary, and the shell command is difficult to understand.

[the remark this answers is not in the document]
tiann: I think check it in code is enough.

kernel/selinux/rules.c
```diff
@@ -170,9 +170,7 @@ static int get_object(char *buf, char __user *user_object, size_t buf_sz,
 // reset avc cache table, otherwise the new rules will not take effect if already denied
 static void reset_avc_cache()
 {
-#if ((KERNEL_VERSION(4, 14, 0) <= LINUX_VERSION_CODE) &&                       \
-     (LINUX_VERSION_CODE < KERNEL_VERSION(4, 14, 163))) ||                     \
-	(LINUX_VERSION_CODE < KERNEL_VERSION(4, 9, 337))
+#if (LINUX_VERSION_CODE < KERNEL_VERSION(5, 10, 0)) && !defined(KSU_COMPAT_USE_SELINUX_STATE)
```
tiann, line 173: we can define a macro standalone, such as:

```
#if (LINUX_VERSION_CODE >= KERNEL_VERSION(5, 10, 0)) || defined(KSU_COMPAT_USE_SELINUX_STATE)
#define KSU_XXXX
#endif 
```

kernel/selinux/Makefile
```diff
@@ -2,6 +2,14 @@ obj-y += selinux.o
...
+  ifeq ($(shell grep -q " current_sid(void)" $(srctree)/security/selinux/include/objsec.h; echo $$?),0)
+  ccflags-y += -DKSU_COMPAT_HAS_CURRENT_SID
+  endif
+  ifeq ($(shell grep -q "struct selinux_state " $(srctree)/security/selinux/include/security.h; echo $$?),0)
+  ccflags-y += -DKSU_COMPAT_USE_SELINUX_STATE
```
tiann, line 10: make this to KSU_COMPAT_HAS_SELINUX_STATE, and we can use

#ifdef KSU_COMPAT_HAS_SELINUX_STATE || version > 5.10
#define KSU_COMPAT_USE_SELINUX_STATE
#endif

tiann (changes requested):

kernel/selinux/Makefile
PaperStrike: Reverted in https://github.com/tiann/KernelSU/pull/402/commits/468120cb099b8489d8a157887431a36a733f675d

kernel/selinux/rules.c
PaperStrike: https://github.com/tiann/KernelSU/pull/402/commits/6bf07bf51559512439290da07af1381faa63de14 done.

tiann (approved):

kernel/selinux/Makefile
PaperStrike: https://github.com/tiann/KernelSU/pull/402/commits/6bf07bf51559512439290da07af1381faa63de14 done.

I put it in `selinux.h`, since both `selinux.c` and `rules.c` have the same check. Let me know if it is inappropriate.

PR #463, opened 2023-05-05
Don't warning build non Kprobes

diphons:
KernelSU work without Kprobes
Tested on kernel xiaomi linux 4.19

### DIFF
--- a/kernel/ksu.c
+++ b/kernel/ksu.c
@@ -30,8 +30,10 @@ int ksu_handle_execveat(int *fd, struct filename **filename_ptr, void *argv,
 					    flags);
 }
 
+#ifdef CONFIG_KPROBES
 extern void ksu_enable_sucompat();
 extern void ksu_enable_ksud();
+#endif
 
 int __init kernelsu_init(void)
 {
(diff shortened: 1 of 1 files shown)


kernel/ksu.c
```diff
@@ -56,7 +58,7 @@ int __init kernelsu_init(void)
 #ifdef CONFIG_KPROBES
 	ksu_enable_sucompat();
 	ksu_enable_ksud();
-#else
+#elif LINUX_VERSION_CODE < KERNEL_VERSION(4, 19, 0)
```
tiann, line 61: Can we remove the condition, and make the `warning` a `info`?

tiann (changes requested):

kernel/ksu.c
diphons: Ya bro.. maybe that's better..
I tried on linux 4.9.x without kprobes, and KSU runs normally too

tiann: 4.9 and 4.14 should also work without kprobes, so we shouldn't warn them, too.

diphons: Ya bro...

tiann:
I will close this since it is stale, welcome to make another PR in any time!

PR #476, opened 2023-05-11
kernel: add ability to exit from safemode without reboot

RooGhz720:
some people got annoyed when they tried to install the module but the manager showed "safemode" and had to reboot, so i added this function, thanks

Usage:
**Volume UP press 3 times to enter safemode**
**Volume Down press 3 times to exit from safemode**


tiann:
I'm not quite sure when to use this. Is it when you boot the device and press the volume down button to enter safe mode and then press the volume up button to exit it? Does it work after the device has been booted?

RooGhz720:
> I'm not quite sure when to use this. Is it when you boot the device and press the volume down button to enter safe mode and then press the volume up button to exit it? Does it work after the device has been booted?

yes work when device booted

RooGhz720:
> I'm not quite sure when to use this. Is it when you boot the device and press the volume down button to enter safe mode and then press the volume up button to exit it? Does it work after the device has been booted?

PR #479, opened 2023-05-12
Update Setup.sh from branch/ksepol

dabao1955:
Instead of forcing kernelsu to be compiled, it is determined whether to compile kernelsu by whether the kprobes option is enabled.


tiann:
We cannot do that because KernelSU can be integrated without KPROBE.

PR #497, opened 2023-05-15
Add option to enable global namespace

Hikari-no-Tenshi:
Disabled by default.
To enable/disable use terminal with root.

To enable:
echo 1 > /sys/module/ksu/parameters/global_namespace_enable

To disable:
echo 0 > /sys/module/ksu/parameters/global_namespace_enable

Value will be reset to 0 (disabled) on reboot.

### DIFF
--- a/kernel/ksu.c
+++ b/kernel/ksu.c
@@ -1,5 +1,6 @@
 #include "linux/fs.h"
 #include "linux/module.h"
+#include "linux/moduleparam.h"
 #include "linux/workqueue.h"
 
 #include "allowlist.h"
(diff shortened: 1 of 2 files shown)


tiann:
What is this for？

Hikari-no-Tenshi:
> What is this for？

This is like Magisk "Mount Namespace Mode" option. If some app doesn't work normally with default namespace (and there is no option in app to switch namespace), you can switch to global namespace to make it work.

tiann (approved):

tiann:
https://github.com/tiann/KernelSU/commit/b80e06256d55e153cd4a642432dff8aa191591cc

Reverted since it can be easily detected by stat /sys/module/ksu/parameters/global_namespace_enable

PR #503, opened 2023-05-16
manager: Update Russian language

TireX228:


tiann:
The feature is in development, please wait it to be stable.

tiann:
Thank you for your contribution, but the feature is in development, it may changes in the future, so we won't accept it now. Welcome to add translation when we finish it!

PR #528, opened 2023-05-18
manager: Update zh-rTW and zh-rHK

Kaitul:


tiann:
Thank you for your contribution, but the feature is in development, it may changes in the future, so we won't accept it now. Welcome to add translation when we finish it!

Kaitul:
Thanks for your reply! I'm watching it's updates.

PR #533, opened 2023-05-19
Need help to build kernelSU for oneplus 7 pro OOS11

khalidaboelmagd:
Hi,

hope you doing will , can you help me to compile kernelSU for oneplus 7 pro with OOS11


tiann:
What is this for? we should always enable safemode, don't make it a kernel parameter.

PR #543, opened 2023-05-21
Fine tuning for Arabic translation.

ahmeda9a:
Done some changes to make the translation Arabic clearer


tiann (approved):

ahmeda9a:
I didn't translate the app profiles because i still didn't use them. Thus, i couldn't translate them properly.
Will translate them when the app profiles gets released

PR #546, opened 2023-05-22
kernel: Fix header inclusion syntax

Forenche:
Follow how upstream does it.


tiann:
This would break VSCode clangd hint, how do you think? @Ylarod

Ylarod:
The clangd auto complete `""` rather than `<>`, so `""` will be more convenient for coding, although it didn't follow the linux coding style

Forenche:
Alright, I just wanted to bring it inline with what upstream does.

PR #557, opened 2023-05-27
Hook improvements

arter97:
Hi @tiann.

Thanks for the great project, I had great fun playing around with it.

This PR mainly tries to further minimize the possible delays caused by KernelSU hooking.

There are 3 major changes:
- Processes with 0 < UID < 2000 are blocked straight-up before going through the allow_list.
I don't see any need for such processes to be interested in root, and this allows returning early before going through a more expensive lookup.
If there's an expected breakage due to this change, I'll remove it. Let me know.
- A simple bitmap 64-bit hash is added.
This speeds up `ksu_is_allow_uid()` by about 38.5%.
- A page-sized (4K) sequential array is added to cache allow_list.
Linux kernel's linked list is efficient enough, but it can be faster.
This speeds up `ksu_is_allow_uid()` by about 3.6x by sacrificing a 4K memory. IMHO, a good trade-off.
Most notably, this reduces the 99.999% result previously from worrying milliseconds scale to microseconds scale.
This also theoretically limits the number of superuser-granted apps to 1024, which is plentiful.

Benchmark results are available at https://docs.google.com/spreadsheets/d/1w_tO1zRLPNMFRer49pL1TQfL6ndEhilRrDU1XFIcWXY/edit?usp=sharing

Other commits are just nice-to-have, each hopefully with enough justification in its commit message.

Thanks!


kernel/allowlist.c
```diff
@@ -67,31 +128,52 @@ bool ksu_allow_uid(uid_t uid, bool allow, bool persist)
 	pr_info("allow_uid: %d, allow: %d", uid, allow);
 
 	list_add_tail(&p->list, &allow_list);
+
+out:
+	if (allow) {
+		allow_list_hash |= HASH_BIT(uid);
+		allow_list_arr[allow_list_pointer++] = uid;
```
tiann, line 135: Could this potentially result in an out-of-bounds write?

kernel/sucompat.c
```diff
@@ -99,25 +99,25 @@ int ksu_handle_execveat_sucompat(int *fd, struct filename **filename_ptr,
...
 	if (IS_ERR(filename)) {
 		return 0;
 	}
 
-	if (!ksu_is_allow_uid(current_uid().val)) {
-		return 0;
-	}
-
-	if (!memcmp(filename->name, su, sizeof(su))) {
+	if (unlikely(!memcmp(filename->name, su, sizeof(su)))) {
```
tiann, line 110: 1. Perhaps we should add some comments here, as it might give the impression that any app can acquire root just by invoking su; however, in reality, apps that are not allowed root access will be denied at the stat and faccessat stages.
2. Would the existence of /system/bin/su in the system lead to root authorization being invalidated, allowing any app to unconditionally obtain root?

tiann:
Thank you for the contribution! I think these are all very good suggestions, but I have some questions:

1. Will there be hash conflicts? If there are hash conflicts, will there be any misjudgments?
2. Is `ksu_is_allow_uid` intentionally placed after judging /system/bin/su in the execve hook? If su exists in the system, it will cause authentication failure (although this is rare).

kernel/allowlist.c
arter97: Good catch, I'll address it.

kernel/sucompat.c
arter97: Ah, didn't review the logic inside the memcmp() check carefully, sorry for that.

We can just return early if memcmp() check fails. I'll reorder the code so that the behavior is consistent with the older one.

arter97:
Thanks for the review @tiann. I'll address those and open up a new PR with a slightly different method soon.

I've managed to squeeze a bit more performance out.

arter97:
New PR opened at #563

PR #562, opened 2023-05-29
change "ブラジル(Portuguese-Brazil)" >> "Português do Brasil"

dabao1955:
I think it might be better? If you are very bothered by this PR, please let me know and I will terminate it.

@AkariOficial


tiann:
#564

PR #570, opened 2023-06-01
setup.sh:change obj-y > obj-$(CONFIG_KSU)

dabao1955:
In my opinion, change "obj-y" to "obj-$(CONFIG_KSU)". I feel that this can decide whether to compile KernelSU instead of the previous forced compilation based on whether the "CONFIG_KSU" option is enabled or not. The source kconfig option has also been added. This allows you to see the KernelSU option in the "Device Drivers" option in menuconfig.


dabao1955:
I think I will should open another Pull Request.

PR #572, opened 2023-06-01
kernel/selinux: Need to force include errno.h sometimes

roynatech2544:
- Seen with Linux 4.14 kernel with error message:

In file included from ../drivers/android/kernelsu/selinux/sepolicy.c:1: In file included from ../drivers/android/kernelsu/selinux/sepolicy.h:6: In file included from ../security/selinux/ss/policydb.h:30: In file included from ../security/selinux/ss/avtab.h:26: ../security/selinux/include/security.h:240:10: error: use of undeclared identifier 'EIDRM'
        return -EIDRM;
                ^
  CC      drivers/base/transport_class.o
  CC      kernel/rcu/update.o
../security/selinux/include/security.h:246:10: error: use of undeclared identifier 'ENOENT'
        return -ENOENT;
                ^

### DIFF
--- a/kernel/selinux/Makefile
+++ b/kernel/selinux/Makefile
@@ -13,4 +13,4 @@ endif
 ccflags-y += -Wno-implicit-function-declaration -Wno-strict-prototypes -Wno-int-conversion
 ccflags-y += -Wno-declaration-after-statement -Wno-unused-function
 ccflags-y += -I$(srctree)/security/selinux -I$(srctree)/security/selinux/include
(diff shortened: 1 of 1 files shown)


roynatech2544:
@tiann is there any concerns in merging this pr?

tiann:
Sorry for the late reply, I believe it will certainly work, but I wonder if there are any better alternatives?

roynatech2544:
Does it seem like a bit dirty fix? Though since that header existed from 3.8 it would be compatible fix

tiann (approved):

PR #576, opened 2023-06-02
setup.sh:change "obj-y"to"obj-$(CONFIG_KSU)"

dabao1955:
reference #570

I think it is maybe necessary.

### DIFF
--- a/kernel/setup.sh
+++ b/kernel/setup.sh
@@ -43,8 +43,7 @@ cd "$GKI_ROOT"
 echo '[+] Add kernel su driver to Makefile'
 
 DRIVER_MAKEFILE=$DRIVER_DIR/Makefile
-grep -q "kernelsu" "$DRIVER_MAKEFILE" || printf "obj-$(CONFIG_KSU) += %s\n" "kernelsu/" >> "$DRIVER_MAKEFILE"
 DRIVER_KCONFIG=$DRIVER_DIR/Kconfig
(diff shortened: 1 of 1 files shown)


dabao1955:
It looks works well

dpkg123:
Should I correct the following? shellcheck doesn't pass if I don't correct it. Although this has no effect on compilation:
change
```
printf "\nobj-$(CONFIG_KSU) += kernelsu/\n"
```
to
```
printf "obj-$(CONFIG_KSU) += %s\n" "kernelsu/"
```

dabao1955:
change `printf "\nobj-$(CONFIG_KSU) += kernelsu/\n"`
to `printf "obj-$(CONFIG_KSU) += %s\n" "kernelsu/"`

Should I correct the following? shellcheck doesn't pass if I don't correct it. Although this has no effect on compilation.

tiann:
Why close it? I think this is a good idea, but the method is not so accurate

dabao1955:
> Why close it? I think this is a good idea, but the method is not so accurate

I think there's a little bit of a problem with that. Especially when KernelSU's Kconfig is inserted into drivers/Kconfigs, it doesn't look so coordinated.

See#582

dabao1955:
> Why close it? I think this is a good idea, but the method is not so accurate



Maybe moving to the main menu is also a good option?

PR #600, opened 2023-06-06
ci: Send media groups directly using the Local Bot API Server

Howard20181:
Should obtaining `api_id` and `api_hash` from https://core.telegram.org/api/obtaining_api_id then create secrets `API_ID` and `API_HASH` first.


scripts/ksubot.py
```diff
@@ -69,27 +69,25 @@ async def main():
     print("[+] Uploading to telegram")
     check_environ()
     print("[+] Files:", sys.argv[1:])
-    bot = telegram.Bot(BOT_TOKEN)
+    bot = telegram.Bot(BOT_TOKEN, base_url='http://127.0.0.1:8088/bot')
```
Ylarod, line 72: check API_ID and fallback to api.telegram.org if this var is not set ?

Ylarod (changes requested):

scripts/ksubot.py
Howard20181: Added.

Ylarod (approved):

Howard20181:
Support for Topics in Groups was added on November 5, 2022, and Action Telegram Bot Server's pre-build api bin was updated on Jan 2022. It was not supported at that time.

Howard20181:
Waiting for response zjns/telegram-bot-server#2, or we can fork and update by ours.

Howard20181:
https://github.com/zjns/telegram-bot-server/issues/2#issuecomment-1592704505
I will try switching to Pyrogram with a larger upload limit (up to 1500M).

Howard20181:
Fixed by f34950723216fd849bd886a79a625ab0838bde1d

PR #631, opened 2023-06-13
Update Russian translation and typo fix

Pegioner:


Pegioner:
Everywhere "App Profile" is mentioned, "Profile" is capitalized, except for <string name="profile" translatable="false">App profile</string> itself. I don't know if it was intended that way, or if it's a typo, but I propose to bring it to a unified form.

tiann (approved):

tiann:
> Everywhere "App Profile" is mentioned, "Profile" is capitalized, except for App profile itself. I don't know if it was intended that way, or if it's a typo, but I propose to bring it to a unified form.

I agree, this is indeed a typo.

tiann (approved):

PR #663, opened 2023-06-19
kernel: fix typo in allowlist.c

eltociear:
creat -> create


tiann:
Although I don't think this is a typo, it doesn't stop me from merging it :)

https://man7.org/linux/man-pages/man3/creat.3p.html

tiann (approved):

PR #692, opened 2023-06-27
sucompat.c:Fix old kernel doesn't have strncpy_from_unsafe_user

Coconutat:
In LINUX KERNEL VERSION 4.4 to 4.13.16,It  doesn't have strncpy_from_unsafe_user, Only have strncpy_from_unsafe.


Coconutat:
> This fix is incorrect, `strncpy_from_unsafe` copies the memory of the kernel space.

So only way is moidfy old kernel?

tiann:
No, you use the wrong API

PR #695, opened 2023-06-27
kernel: print the current KernelSU version at build time

Aquarius223:
* For example:
   - KernelSU version: 11055

(cherry picked from commit 46aa8e7a32e9c84dcd9e7f1ba4e16109949e6bf1) Change-Id: Iaaf4cd4100c9e344a6df6ea73fa74a347b81297b


tiann:
we calculate it https://github.com/tiann/KernelSU/blob/main/kernel/ksu.h#L12, i think it is a bad idea to calc it twice.

PR #696, opened 2023-06-28
kernel: Modify KERNEL_SU_VERSION logic and behavior

Aquarius223:
This patch modifies the following:
- Move the version addition logic in ksu.h to Makefile processing
- Print the current version number of KernelSU during make build

build test (normal):
 buildlog:
  HOSTCC  scripts/basic/bin2c Using .. as source for kernel -- KernelSU version: 11055
  CHK     include/generated/utsrelease.h
  UPD     include/generated/utsrelease.h

 KernelSU software:
  KernelSU Working Version: 11055 (v) Superusers: 0 Modules: 0

build test (missing .git file):
 buildlog:
  HOSTCC  scripts/basic/bin2c Using .. as source for kernel ../drivers/kernelsu/Makefile:23: "KSU_GIT_VERSION not defined! It is better to make KernelSU a git submodule!"
  CHK     include/generated/utsrelease.h
  CHK     scripts/mod/devicetable-offsets.h

 KernelSU software:
  KernelSU Working Version: 16 (v) Superusers: 0 Modules: 0

  The current kernel version 16 is too low for the manager to function properly. Please upgrade to version 10977 or higher!

(cherry picked from commit 0c38a1614a77d80de752aba20908e3f9d21660a8) Change-Id: I570f2ee33db224e1a36770f847137f290ba9bcfd

### DIFF
--- a/kernel/Makefile
+++ b/kernel/Makefile
@@ -15,7 +15,13 @@ obj-y += selinux/
 # .git is a text file while the module is imported by 'git submodule add'.
 ifeq ($(shell test -e $(srctree)/$(src)/../.git; echo $$?),0)
 KSU_GIT_VERSION := $(shell cd $(srctree)/$(src); /usr/bin/env PATH="$$PATH":/usr/bin:/usr/local/bin git rev-list --count HEAD)
-ccflags-y += -DKSU_GIT_VERSION=$(KSU_GIT_VERSION)
+# ksu_version: major * 10000 + git version + 200 for historical reasons
+$(eval KSU_VERSION=$(shell expr 10000 + $(KSU_GIT_VERSION) + 200))
+$(info -- KernelSU version: $(KSU_VERSION))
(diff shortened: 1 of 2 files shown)


tiann:
Does the `#pragma message` work? maybe we can show the info in ksu.h

Aquarius223:
> 



> Does the `#pragma message` work? maybe we can show the info in ksu.h

The message is valid and displays the text as it should (after deliberately mv .git .gitt)
I think the define in the header file should be as clear as possible, and I think it is also a good effect to pass the parameters to define the value through the makefile

tiann:
Doing calculations in Makefile means we have to deal with a variety of environments, and I'm not sure if this is supported in all shells.

Aquarius223:
> Doing calculations in Makefile means we have to deal with a variety of environments, and I'm not sure if this is supported in all shells.

The only consideration in the shell is `expr` which comes pre-installed on modern systems, including Android (system/bin/expr)

tiann (approved):

PR #732, opened 2023-07-08
kernel:selinux:sepolicy.c:Add support for Huawei HarmonyOS (based on EMUI 10) kernel

Coconutat:
I want to support Huawei HarmonyOS (based on EMUI 10) kernel.  
Huawei has changed the code as follows:
```
#ifdef CONFIG_HKIP_SELINUX_PROT
static inline void ebitmap_init(struct ebitmap *e, const bool protectable)
{
	memset(e, 0, sizeof(*e));
	e->protectable = protectable;
}
#else
static inline void ebitmap_init(struct ebitmap *e, const bool protectable)
{
	(void)protectable;
	memset(e, 0, sizeof(*e));
}
#endif
```

So I disable CONFIG_HKIP_SELINUX_PROT,then just use this part:  
```
static inline void ebitmap_init(struct ebitmap *e, const bool protectable)
{
	(void)protectable;
	memset(e, 0, sizeof(*e));
}
```  
Then just modify this KSU part:
```
ebitmap_init(flex_array_get(db->type_attr_map_array, value - 1));
```
The Kernel will be compiled successful.

### DIFF
--- a/kernel/selinux/sepolicy.c
+++ b/kernel/selinux/sepolicy.c
@@ -12,12 +12,20 @@
 /*
  * Adapt to Huawei HISI kernel without affecting other kernels ,
  * Huawei Hisi Kernel EBITMAP Enable or Disable Flag ,
- * From ss/ebitmap.h
+ * From ss/ebitmap.h.
+ * If the kernel is designed for HarmonyOS(Based EMUI 10),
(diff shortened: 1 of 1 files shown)


Coconutat:
Not stable.Closed.

PR #753, opened 2023-07-12
Add the build debug manager workflow

chen2021-web:


tiann:
Please reuse the workflow

PR #754, opened 2023-07-12
change AkariOficial to whyakari

dabao1955:
AkariOficial has been renamed to whyakari. Entering https://github.com/AkariOficial directly will result in 404.


tiann:
I think we should remove the repo because it doesn't contain any source code.

dabao1955:
Yes,you are right.

Issue #736, opened 2023-07-09
su 命令参数解析不正常


Mufanc:
如图，版本 v0.6.0 (10981)

![image](https://github.com/tiann/KernelSU/assets/47652878/15a811f5-10b0-4924-b90b-77a4b2fdd5a2)

tiann:
https://github.com/tiann/KernelSU/commit/df9cf61575dcdfc0a160929bd8d85210d61464cb

PR #808, opened 2023-07-26
ksud: try to fix issue #736

TinyHai:
I just tried to fix it. To be honest, I am not sure if it will work.


tiann:
This won't work

PR #816, opened 2023-07-31
Fix the directory pathname while moving directories

AzukiAtsui:
Fix the duplicate directory name error that occurs when the destination directory already exists while moving folders.

### DIFF
--- a/userspace/ksud/src/installer.sh
+++ b/userspace/ksud/src/installer.sh
@@ -301,6 +301,20 @@ is_legacy_script() {
   return $?
 }
 
+# find_mv [source_directory] [destination_directory]
+find_mv() {
+    for file in $(find "$1" -type f); do
+        # Get the sub directory of the file.
+        sub_dir=$(echo "${file%/*}" | sed "s|$1||")
+        # Create the new directory, if it doesn't already exist.
+        mkdir -p "$2$sub_dir"
+        # Move the file to the new directory.
+        mv -f "$file" "$2$sub_dir"
+    done
+    # Clean old directory.
+    rm -r "$1"
+}
+
 handle_partition() {
     # if /system/vendor is a symlink, we need to move it out of $MODPATH/system, otherwise it will be overlayed
     # if /system/vendor is a normal directory, it is ok to overlay it and we don't need to overlay it separately.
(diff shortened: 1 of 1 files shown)


tiann:
Under what circumstances would a user create a vendor/product file in the module root directory? If a user creates it unknowingly and the file gets overwritten, the issue would still exist.

AzukiAtsui:
I have noticed the WARNING [`DO NOT MANUALLY CREATE OR MODIFY`](https://kernelsu.org/guide/module.html#kernelsu-modules).
However, `install_module` creates vendor/product/system_ext files in MODPATH during installation if any user uses variable REMOVE or REPLACE.
Currently the lines of `# Handle remove files` are superior of `handle_partition` lines in installer.sh, so here comes the bug: modified files in MODPATH/system/system_ext which are supposed to overlay would be moved to MODPATH/system_ext/system_ext then the overlay path wrongs too.

tiann:
I think I understand what you mean by:

If the user wants to REMOVE `/vendor/priv-app/A`, and `/system/vendor` is a symlink to `/vendor`; if it writes:

`REMOVE=/vendor/priv-app/A` is normal, but if it uses

`REMOVE=/system/vendor/priv-app/A` will result in wrong path; is it so?

AzukiAtsui:
Sorry I did not clearly comment. Let me try describing more.

If the user wants to REMOVE `/vendor/priv-app/A`, and `/system/vendor` is a symlink to `/vendor`; and MODULE ZIP FILE contains `/system/vendor/priv-app/B` to REPLACE `/vendor/priv-app/B`;
if he writes:

`REMOVE=/vendor/priv-app/A`
    - `/vendor/priv-app/A` will be REMOVEd normally;
    - `/vendor/priv-app/B` will NOT be REPLACEd, it will be in `/vendor/vendor/priv-app/B`.
    - At this case, `$MODPATH/vendor` created by REMOVE exists, then `mv $MODPATHE/system/vendor $MODPATH/vendor` will result in wrong path.

`REMOVE=/system/vendor/priv-app/A`
    - `/vendor/priv-app/A` will be REMOVEd normally;
    - `/vendor/priv-app/B` will be PEPLACEd normally.
    - At this case, in module root path, all files moved from `/system/vendor` to `/vendor` without problem(existance of destination directory) and everything works well.

tiann:
Thank you, i understand it.

tiann (approved):

tiann:
I'll revert this since it causes other bugs:

1. `REPLACE=/system/product/app/A` would fail, because there isn't any file in directory, only dir, so `find "$1" -type f` would find nothing.
2. `REMOVE=/system/product/app/A` would fail, because `find "$1" -type f` would skip character devices.

PR #825, opened 2023-08-02
manager: Corrected "Kernel version" and "KernelSU version" string

Aquarius223:
* Avoid misunderstanding "The current kernel version %d is too low" Many people misunderstand that the version of the Linux kernel is too low, so the version of KernelSU is low and they go looking for a kernel with an upgraded version of the Linux kernel.

 Correct the string of require_kernel_version to avoid misunderstanding,
 Now let them look for the updated KernelSU version of the kernel.

(cherry picked from commit dcc4ad10f81b1531b551b2b44a3b9cbdc0e489e9) Change-Id: I4373b006d45ed98e02effa4556e8a9e8c0b70f14


tiann:
You need to delete all other translation to make them re-translate it.

tiann (approved):

PR #826, opened 2023-08-03
Update readonly strings on Weblate (Vietnamese)

cachiusa:


tiann:
Please edit in https://hosted.weblate.org/projects/kernelsu/, otherwise it may conflict

cachiusa:
Can you make the 'app profile' string editable on Weblate? It won't let me change that string

tiann:
No, don't translate this.

PR #827, opened 2023-08-03
Add WSA Kernel `5.15.104.2`

s1204IT:
Remove `5.15.94.4` as soon as WSA 2306 is removed from the release.


s1204IT:
Activate the PR again when WSA 2307 is published in the public WIF.

tiann (approved):

PR #866, opened 2023-08-14
feat(kernel): Add Kernel 5.10.187

pixincreate:
- Bluejay running GrapheneOS is on Kernel version `5.10.187`.  
- Tried to flash other versions `5.10.xxx` and all went into bootloop. Unsure how this version would go

### DIFF
--- a/.github/workflows/build-debug-kernel.yml
+++ b/.github/workflows/build-debug-kernel.yml
@@ -17,8 +17,8 @@ jobs:
       matrix:
         include:
           - version: "5.10"
-            sub_level: 177
-            os_patch_level: 2023-06
+            sub_level: 187
+            os_patch_level: 2023-08
           - version: "5.15"
             sub_level: 119
             os_patch_level: 2023-09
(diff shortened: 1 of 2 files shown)


pixincreate:
How to know why it failed? I'm unable to understand that..

tiann:
Google has not published it: https://source.android.com/docs/core/architecture/kernel/gki-android13-5_10-release-builds#launch-releases

pixincreate:
Got it. I never thought about this. Until Google releases the kernel, I cannot use this unless I manually patch and install.  
Is this the reason why the device went into bootloop or just because I messed with kernel that GrapheneOS provided with the ROM (they do their own kernel hardening)

pixincreate:
@tiann, do you think this would work on GrapheneOS? Asking because GrapheneOS heavily modifies Kernel to do hardening of the OS. Have heard that this would break ABI. Last time when I tried installing KernelSU, it straight went into bootloop..

tiann:
> @tiann, do you think this would work on GrapheneOS? Asking because GrapheneOS heavily modifies Kernel to do hardening of the OS. Have heard that this would break ABI. Last time when I tried installing KernelSU, it straight went into bootloop..

I don't know.

tiann (approved):

PR #877, opened 2023-08-16
Fix kernel 5.15.119

zhantech:


tiann:
https://github.com/tiann/KernelSU/pull/866#issuecomment-1678554198

PR #932, opened 2023-09-04
Update unofficially-support-devices.md

DrBrown28:
This is the kernel Source For Red Magic 5s/5g with KSU support in LineageOS ROM


dabao1955:
Why not to change repos.json？

PR #958, opened 2023-09-15
Copy one extra byte from userspace filename

### LINKED ISSUE #957: Incorrect syscall results on filenames that begins with /system/bin/su
**Describe the bug**

Incorrect results are given to userland when kernelsu handles filenames that begins with `/system/bin/su`

**To Reproduce**
Steps to reproduce the behaviour:
1. Run `adb shell` (Super user is granted to shell)
2. In adb shell, run `suadfadg`. 
3. I got `/system/bin/sh: /system/bin/suadfadg: No such file or directory` 

**Expected behaviour**

```
/system/bin/sh: fdsfsf: inaccessible or not found
```

**Smartphone (please complete the following information):**
Not relevant. All devices are affected.

**Additional context**

This is because kernelsu rewrites `su->sh` because the prefix of `/system/bin/suadfadg` matches `/system/bin/su`.

I will open a PR to fix it.

kxxt:
Otherwise we will rewrite paths for filenames that begins with `/system/bin/su`.

This fix copies one extra byte from userspace filename so that when we encounter filenames like `/system/bin/suasf`,
`/system/bin/su\0` gets compared with `/system/bin/sua`, which correctly prevents the `su -> sh` path rewriting.

Close #957

### DIFF
--- a/kernel/sucompat.c
+++ b/kernel/sucompat.c
@@ -48,7 +48,7 @@ int ksu_handle_faccessat(int *dfd, const char __user **filename_user, int *mode,
 		return 0;
 	}
 
-	char path[sizeof(su)];
+	char path[sizeof(su) + 1];
 	memset(path, 0, sizeof(path));
 	ksu_strncpy_from_user_nofault(path, *filename_user, sizeof(path));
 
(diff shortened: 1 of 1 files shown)


kernel/sucompat.c
```diff
@@ -48,7 +48,7 @@ int ksu_handle_faccessat(int *dfd, const char __user **filename_user, int *mode,
 		return 0;
 	}
 
-	char path[sizeof(su)];
+	char path[sizeof(su) + 1];
 	memset(path, 0, sizeof(path));
 	ksu_strncpy_from_user_nofault(path, *filename_user, sizeof(path));
```
tiann, line 53: Maybe we should use `copy_from_user_nofault` instead of `strncpy_from_user_nofault`, but this should also fix it.

tiann (approved):

PR #966, opened 2023-09-17
Update unofficially-support-devices.md

hc841:


tiann:
you should modify this: https://github.com/tiann/KernelSU/blob/main/website/docs/repos.json or open an issue: https://github.com/tiann/KernelSU/issues/new?assignees=&labels=add-device&projects=&template=add_device.yml&title=%5BAdd+Device%5D%3A+

PR #973, opened 2023-09-19
kernel: Adapt to low version Android init process

longhuan1999:
1. Adapt to low version Android init process
2. Add stop hook output
3. Fix output with missing line breaks

### DIFF
--- a/kernel/ksud.c
+++ b/kernel/ksud.c
@@ -103,44 +103,6 @@ static const char __user *get_user_arg_ptr(struct user_arg_ptr argv, int nr)
 	return native;
 }
 
-static const bool get_env_bool(struct user_arg_ptr envp, const char *name) {
-	const char __user *const __user *env;
-
-	// Traverse the list of environment variables
-    for (env = envp.ptr.native; *env != NULL; env++) {
-        const char __user *entry = *env;
-        char buf[256];
-        
-        // Reading environment variable strings from user space
-        if (ksu_strncpy_from_user_nofault(buf, entry, sizeof(buf)) < 0)
-            continue;
-        
-        // Parsing environment variable names and values
-        char *env_name = buf;
-        char *env_value = strchr(buf, '=');
-        if (env_value == NULL)
-            continue;
-        
-        // Replace equal sign with string terminator
-        *env_value = '\0';
-        env_value++;
-        
-        // pr_info("env %s=%s\n", env_name, env_value);
-        // Check if the environment variable names are matching
-        if (strcmp(env_name, name) == 0) {
-            // If matched, returns a Boolean value based on the environment variable's value
-            if (strcmp(env_value, "1") == 0 || strcmp(env_value, "true") == 0)
-                return true;
-            else
-                return false;
-        }
-    }
-    
-    // If the specified environment variable is not found, return the default value
-	pr_info("not found env %s\n", name);
-    return false;
-}
-
 /*
  * count() counts the number of strings in array ARGV.
  */
@@ -178,7 +140,7 @@ static int __maybe_unused count(struct user_arg_ptr argv, int max)
 
 // the call from execve_handler_pre won't provided correct value for __never_use_argument, use them after fix execve_handler_pre, keeping them for consistence for manually patched code
 int ksu_handle_execveat_ksud(int *fd, struct filename **filename_ptr,
-			     struct user_arg_ptr *argv, struct user_arg_ptr *envp, int *__never_use_flags)
+				struct user_arg_ptr *argv, struct user_arg_ptr *envp, int *__never_use_flags)
 {
 #ifndef CONFIG_KPROBES
 	if (!ksu_execveat_hook) {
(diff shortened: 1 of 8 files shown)


kernel/ksud.c
```diff
@@ -103,6 +103,44 @@ static const char __user *get_user_arg_ptr(struct user_arg_ptr argv, int nr)
...
+        
+        // Replace equal sign with string terminator
+        *env_value = '\0';
+        env_value++;
+        
+        // pr_info("env %s=%s\n", env_name, env_value);
+        // Check if the environment variable names are matching
+        if (strcmp(env_name, name) == 0) {
+            // If matched, returns a Boolean value based on the environment variable's value
+            if (strcmp(env_value, "1") == 0 || strcmp(env_value, "true") == 0)
```
tiann, line 132:
```suggestion
return strcmp(env_value, "1") == 0 || strcmp(env_value, "true") == 0;
```

kernel/ksud.c
```diff
@@ -163,24 +205,37 @@ int ksu_handle_execveat_ksud(int *fd, struct filename **filename_ptr,
 	}
 
 	if (unlikely(!memcmp(filename->name, system_bin_init,
-		    sizeof(system_bin_init) - 1))) {
```
tiann, line 166: Don't bother the logic of Android 10+. We don't want to introduce new bugs for just adapting Android 10-

tiann: You should make it a single condition block, don't modify code of Android 10+. If Android 6-7 needs to modify it, please drop.

tiann (changes requested):

tiann (changes requested):

kernel/ksud.c
```diff
@@ -103,6 +103,41 @@ static const char __user *get_user_arg_ptr(struct user_arg_ptr argv, int nr)
 	return native;
 }
 
+static const bool get_env_bool(struct user_arg_ptr envp, const char *name) {
```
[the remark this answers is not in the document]
longhuan1999: I think so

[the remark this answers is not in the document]
tiann: So can we just use `get_user_arg_ptr`, i think it is more stable than yours.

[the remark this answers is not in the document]
longhuan1999: Actually, I don't know how to do it. I have not studied C++ and kernel development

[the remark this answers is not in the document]
tiann: 1. First, you can use `count(*envp, MAX_ARG_STRINGS)` to get the env count.
2. Second, you can use `get_user_arg_ptr(envp, i)` to get every env variable.
3. And then you can check every env variable to see if it hit your condition.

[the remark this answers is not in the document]
longhuan1999: ok

kernel/ksud.c
```diff
@@ -183,10 +187,63 @@ int ksu_handle_execveat_ksud(int *fd, struct filename **filename_ptr,
...
+			}
+		} else if (argc == 1 && !init_second_stage_executed) {
+			/* This applies to versions between Android 8 ~ 9  */
+			int envc = count(*envp, MAX_ARG_STRINGS);
+			if (envc > 0) {
+				int n;
+				for (n = 1; n <= envc; n++) {
+					const char __user *p = get_user_arg_ptr(*envp, n);
+					if (p && !IS_ERR(p)) {
+						char env[256];
```
tiann, line 219: Why copy twice?

kernel/ksud.c
```diff
@@ -183,10 +187,63 @@ int ksu_handle_execveat_ksud(int *fd, struct filename **filename_ptr,
...
+				pr_err("/init parse args err!\n");
+			}
+		} else if (argc == 1 && !init_second_stage_executed) {
+			/* This applies to versions between Android 8 ~ 9  */
+			int envc = count(*envp, MAX_ARG_STRINGS);
+			if (envc > 0) {
+				int n;
+				for (n = 1; n <= envc; n++) {
+					const char __user *p = get_user_arg_ptr(*envp, n);
+					if (p && !IS_ERR(p)) {
```
tiann, line 218:
```suggestion
					if (!p || IS_ERR(p)) {
					    continue;
					}
```

tiann (changes requested):

kernel/ksud.c
```diff
@@ -183,10 +187,65 @@ int ksu_handle_execveat_ksud(int *fd, struct filename **filename_ptr,
...
+			/* This applies to versions between Android 8 ~ 9  */
+			int envc = count(*envp, MAX_ARG_STRINGS);
+			if (envc > 0) {
+				int n;
+				for (n = 1; n <= envc; n++) {
+					const char __user *p = get_user_arg_ptr(*envp, n);
+					if (!p || IS_ERR(p)) {
+					    continue;
+					}
+					if (p && !IS_ERR(p)) {
```
tiann, line 221: we can delete this condition now.

tiann (changes requested):

kernel/ksud.c
longhuan1999: No need (p && !IS_ERR(p)) and (!p || IS_ERR(p))? 
What do they mean? 
Are they same with (ksu_strncpy_from_user_nofault(env, p, sizeof(env)) < 0)?

tiann: Because we have
```
if (!p || IS_ERR(p)) {
	continue;
}
```

longhuan1999: Is that all right now?

tiann: It looks good to me now. One final question: Does it work on your device? I have no device to test.

longhuan1999: I just tested it and it works properly on my device.

tiann (approved):

PR #1032, opened 2023-10-12
ci: Add android13-5.15-lts

zhantech:


tiann:
No device uses tls kernel

PR #1057, opened 2023-10-19
Add Android 14 kernel ci

tiann:


amteza:
Just checked your new actions to build for Android 14, in my case I used the next command to compile the kernel:
`CCACHE="/usr/bin/ccache" tools/bazel build --config=fast --subcommands //common:kernel_aarch64_dist`

And the next OUTDIR in the action to package it:
`OUTDIR=$GITHUB_WORKSPACE/android-kernel/bazel-out/k8-fastbuild/bin/common/kernel_aarch64`

It built but it doesn't run properly in the device. I might be missing something else.

Issue #1050, opened 2023-10-16
[Kernel 4.4] Kernelsu freeze on installing modules with KSU app


Aryanxfa:
**Describe the bug**
Any module installation gets stuck at "Mounting Image" and the device freezes. Need to perform a force reboot to restore functionality. The thing to note is that turning selinux permissive allows for proper installation of modules. 

**To Reproduce**
Installing any module causes that issue. 

**Expected behaviour**
The module installation should either give a clear failure or a success without freezing the device

**Screenshots**
![photo_6154388858008091188_y](https://github.com/tiann/KernelSU/assets/26765117/e3cc89f9-a84b-4bf1-af7a-130547a7458f)

**Smartphone (please complete the following information):**
 - Device: Samsung Galaxy A20 ( same on a10,a20e,a30,a30s,a40 )
 - OS: Cherish Android 13 v4.12 ( tested with lineage 20.0, same issue )

**Additional context**
As the problem is solved after setting selinux to permissive. i have pulled some ksu app logs, module installation logs and avc_msg
attaching them here for your reference. i am astonished as to why u:r:su:s0 is getting denials.

`type=1400 audit(1695268737.027:3064): avc:  denied  { ioctl } for  pid=19800 comm="e2fsck" path="/data/adb/ksu/update_tmp.img" dev="mmcblk0p32" ino=785282 ioctlcmd=127c scontext=u:r:su:s0 tcontext=u:object_r:adb_data_file:s0 tclass=file permissive=1
`

Kernel source:- https://github.com/eurekadevelopment/Eureka-Kernel-Exynos7885-Q-R-S
KernelSu commit :- https://github.com/eurekadevelopment/Eureka-Kernel-Exynos7885-Q-R-S/commit/41d4196ec2fe6543893b378abe75dba4eb5ceef9
Device tree:- https://github.com/eurekadevelopment/android_device_samsung_exynos7885/

[KernelSU_bugreport_2023-10-16_22_41.tar.gz](https://github.com/tiann/KernelSU/files/12918933/KernelSU_bugreport_2023-10-16_22_41.tar.gz)

[KernelSU_install_log_2023-10-16-22-40-09.log](https://github.com/tiann/KernelSU/files/12918938/KernelSU_install_log_2023-10-16-22-40-09.log)

[avc_msg.txt](https://github.com/tiann/KernelSU/files/12919001/avc_msg.txt)

Any suggestion / guidance in the right direction would be extremely helpful

Thanks!

[...]

PR #1059, opened 2023-10-19
KernelSU: selinux: fix Modules install stuck on kernel 4.4

kevios12:
This fixes that the Module Installer is stuck when installing any Modules on Kernel 4.4


tiann:
https://github.com/tiann/KernelSU/issues/1050#issuecomment-1766762513  Does this miss some rules?

tiann (changes requested):
https://github.com/tiann/KernelSU/issues/1050#issuecomment-1766762513  Does this miss some rules?

kevios12:
We added the missing rules to our ROM. Exactly that in the comment

tiann (approved):

PR #1121, opened 2023-11-07
[Remove Device]: Remove a repository that has been archived

Citrinae-Lime:
https://github.com/LeviMarvin/android_kernel_xiaomi_alioth
Has been archived by the owner on Oct 28, 2023. It was been unactive since April, 2023.

### DIFF
--- a/website/docs/repos.json
+++ b/website/docs/repos.json
@@ -147,13 +147,6 @@
         "devices": "OPPO Reno Ace (OP4A89)"
     },
     {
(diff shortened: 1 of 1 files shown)


tiann (approved):

tiann:
I think "inactive" is not the reason to remove it, the kernel needn't to update frequently. If you can confirm that it doesn't work on any version of Android12/13/14, we can delete it.

PR #1128, opened 2023-11-12
Check if contain module id

Howard20181:


userspace/ksud/src/module.rs
```diff
@@ -663,7 +663,7 @@ fn _list_modules(path: &str) -> Vec<HashMap<String, String>> {
                 module_prop_map.insert(k, v);
             });
 
-        if module_prop_map["id"].is_empty() {
+        if !module_prop_map.contains_key("id") || module_prop_map["id"].is_empty() {
```
tiann, line 666: `module_prop_map.get("id").is_none()`

tiann (approved):

userspace/ksud/src/module.rs
tiann: I prefer `!module_prop_map.filter(|s| !s.is_empty()).is_some()` but this expression is more readable.

PR #1134, opened 2023-11-13
templates: Update templates and try to get rid with `CAP_SYS_ADMIN`

Rem01Gaming:
- The use of CAP_SYS_ADMIN can be avoided in Kernel Manager, but cannot be avoided in Root Explorer because it's needed for mounting RW/RO.
- Capabilities adjustment
- Fix template typo

### DIFF
--- a/website/docs/public/templates/kernelmanager.root
+++ b/website/docs/public/templates/kernelmanager.root
@@ -6,14 +6,14 @@
     "uid":0,
     "gid":0,
     "groups":[
-        "ROOT"
+        "ROOT",
+        "READPROC"
     ],
     "capabilities":[
         "CAP_SYS_MODULE",
         "CAP_SYS_NICE",
         "CAP_SYS_RESOURCE",
         "CAP_KILL",
-        "CAP_SYS_PTRACE",
         "CAP_SYSLOG",
         "CAP_PERFMON"
     ],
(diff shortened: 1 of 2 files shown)


tiann:
What is the usage of `CAP_SYS_PTRACE`?

Rem01Gaming:
Needed for see all running process

tiann:
Add it to group `read_proc` is enough.

Rem01Gaming:
Yes, I'm also thinking the same but FKM's perfmon doesn't shows all processes without it.
Attachment : 
[1]: FKM With `CAP_SYS_PTRACE` Screenshot_20231113-100349_Franco Kernel Manager
[2]: FKM Without `CAP_SYS_PTRACE` 
Screenshot_20231113-100438_Franco Kernel Manager

Rem01Gaming:
Yes I'm also thinking the same thing but FKM's perfmon process doesn't displays all processes.

FKM with `CAP_SYS_PTRACE`
Screenshot_20231113-100349_Franco Kernel Manager
FKM without `CAP_SYS_PTRACE`

tiann:
`groups: ["ROOT", "READPROC"]` should be ok

Rem01Gaming:
> Yes I'm also thinking the same thing but FKM's perfmon process doesn't displays all processes.

It's working now
Screenshot_20231113-121715_Franco Kernel Manager

tiann (approved):

PR #1144, opened 2023-11-14
Update repos.json

uurcan7:
MOE officially no longer supports Android 11.

### DIFF
--- a/website/docs/repos.json
+++ b/website/docs/repos.json
@@ -525,13 +525,6 @@
         "devices": "Samsung Galaxy A70 ( A705MN / A705FN )"
     },
     {
-        "maintainer": "whyakari",
-        "maintainer_link": "https://github.com/whyakari",
-        "kernel_name": "android_kernel_xiaomi_ginkgo",
-        "kernel_link": "https://github.com/whyakari/android_kernel_xiaomi_ginkgo",
-        "devices": "non-GKI ginkgo"
-    },
-    {
         "maintainer": "wulan17",
         "maintainer_link": "https://github.com/wulan17",
         "kernel_name": "android_kernel_xiaomi_mt6765",


tiann:
It just support KernelSU 0.7.1 two days ago, why remove it?

uurcan7:
It says on the "[Unofficially Supported Devices | KernelSU](https://kernelsu.org/guide/unofficially-support-devices.html)" page:
"WARNING
In this page, there are kernels for non-GKI devices supporting KernelSU maintained by other developers."
But moe now:
resim
In other words, since it now supports Android 12 and above, it is a supported kernel and is no longer non-GKI.
If you want, you can check the download link from the Moe producer's telegram channel.

tiann:
Android 12 != GKI

uurcan7:
> Android 12 != GKI

As I said, the page I mentioned only contains kernels that support non-GKI Android versions. and now moe only supports gki kernels. In other words, it cannot be hosted on the page I mentioned because it does not support non-GKI Android versions.

tiann:
Please read the KernelSU docs carefully, you just don't know what GKI means.

uurcan7:
The page in question only contains NON-GKI android versions, that is, kernels that support versions below 12 as UNOFFICIALLY SUPPORTED. However, since moe no longer supports NON-GKI android versions below 12, it cannot remain on the UNOFFICIAL SUPPORTED page. Because NON-GKI means below Android 12. However, Moe now only supports 12 and above. It's pathetic that you don't understand this.

tiann:
Android version is nothing to do with kernel version, do you understand？A device with Android 12 doesn't means it is GKI kernel. Only kernel 5.10 could be GKI.

I told you to read the docs alreay, so stop abuse issue here.

PR #1200, opened 2023-12-10
german fix

Henry2o1o:


tiann:
thank you, but please use https://hosted.weblate.org/projects/kernelsu/manager to contribute the translation, otherwise it may conflict and is hard to maintain.

PR #1246, opened 2024-01-04
Correctly handle is_ksu_domain, close #972

### LINKED ISSUE #972: su command failing in addon.d-v2+ scripts
**Describe the bug**
KernelSU does not seem to have a way to allow su commands to function when executed from an addon.d script.

**Steps to reproduce**
1. Use an A/B device and custom ROM with addon.d support (e.g. Lineage)
2. Use root to add an addon.d-v2+ (i.e. A/B) script which uses su to /system/addon.d and set `chmod 755` it (e.g. 97-dumpimages.sh - unzip from attached [97-dumpimages.sh.zip](https://github.com/tiann/KernelSU/files/12651889/97-dumpimages.sh.zip))
3. Accept a ROM A/B OTA, let it complete then reboot as instructed
4. See that the script did not function (e.g. no partition dumps were created in /sdcard/Download)

**Log**
I piped output from the su command in the script, which confirms it fails with the following:
```
/postinstall/tmp/addon.d/97-dumpimages.sh[135]: su: inaccessible or not found
```

**Expected behaviour**
KernelSU has a com.android.shell entry in the default list to allow `adb shell` su commands, and should have something similar to allow su commands from the update_engine/backuptool so that addon.d scripts using su can function like they would with Magisk.

I tried enabling root for com.android.shell, and other system packages like android, systemui, Lineage's updater, etc, with no luck, but there must be something that KernelSU can do to allow it.

**Smartphone info**
 - Device: OnePlus 7T
 - OS: Lineage 20 (Android 13)
 - Version: KernelSU 11191 + Manager v0.6.8 (11238)

Ylarod:
fix #972

### DIFF
--- a/kernel/selinux/selinux.c
+++ b/kernel/selinux/selinux.c
@@ -99,24 +99,15 @@ static inline u32 current_sid(void)
 }
 #endif
 
-static u32 ksu_sid = 0;
-
 bool is_ksu_domain()
 {
-	int error;
-	u32 sid;
-	if (!ksu_sid) {
-		error = security_secctx_to_secid(KERNEL_SU_DOMAIN, strlen(KERNEL_SU_DOMAIN), &sid);
-		if (error) {
-			pr_info("security_secctx_to_secid %s -> sid: %d, error: %d\n",
-				KERNEL_SU_DOMAIN, sid, error);
-		}
-		if (!error) {
-			if (!ksu_sid)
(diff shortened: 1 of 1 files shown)


tiann (approved):

osm0sis:
Thanks so much! I'll test it out the next chance I get! 😁

PR #1267, opened 2024-01-07
website: update Galaxy A52s/M52 repo link

BlackMesa123:


tiann:
Please don't use the tag as link name, this would change now and then

tiann (approved):

PR #1274, opened 2024-01-08
website/docs/guide: Enable KernelSU support (CONFIG_KSU)

whyakari:
### enable CONFIG_KSU on the website documentation.

Added support for KernelSU by enabling CONFIG_KSU in the kernel defconfig located at `arch/arm64/configs/vendor/your_defconfig`. Please ensure to verify the correct location of the file. Note that on some devices, the defconfig may be located in `arch/arm64/configs`.


tiann:
KernelSU doesn't support 32bit kernel

whyakari:
> KernelSU doesn't support 32bit kernel

really, sorry. adjusted.

whyakari:
Could you add CONFIG_KSU in the website documentation? without defining `#ifdef` and `#endif` between the hooks, it gives a compilation error, so it's good to add this in the documentation, as it helps other developers

tiann:
> Could you add CONFIG_KSU in the website documentation? without defining `#ifdef` and `#endif` between the hooks, it gives a compilation error, so it's good to add this in the documentation, as it helps other developers

This is a good idea, just delete the description of 32bit is ok for this PR

tiann (approved):

PR #1330, opened 2024-02-01
Added shizuku.root template

JohnRTitor:
Shizuku only needs DAC_OVERRIDE to execute the script to start the Shizuku service.
Also added bengali translation for the name and description.

### DIFF
--- a/website/docs/public/templates/shizuku.root
+++ b/website/docs/public/templates/shizuku.root
@@ -6,7 +6,7 @@
     "uid":0,
     "gid":0,
     "groups":[
-        "ROOT"
+        "SHELL"
     ],
     "capabilities":[
         "CAP_DAC_OVERRIDE"
--- a/website/docs/public/templates/shizuku.root
+++ b/website/docs/public/templates/shizuku.root
@@ -1,0 +1,22 @@
+{
+    "id":"shizuku.root",
+    "name":"Shizuku Service",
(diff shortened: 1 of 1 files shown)


JohnRTitor:
Excuse me? Is there a reason it is closed?

yujincheng08:
I think we can only gain shell permission for it.

JohnRTitor:
You are right. It only needs to be part of the `shell` group. But it needs DAC_OVERRIDE permission to execute shell scripts. Else we'll get a permission error like this: 
Screenshot_20240201-220033_Shizuku


With the DAC_OVERRIDE and shell group, it can successfully start Shizuku service.

Screenshot_20240201-220712_KernelSU


You can test on your own if you want.

JohnRTitor:
Can I update the commit and create another pull request?

yujincheng08:
I reopened it. You can directly update it.

JohnRTitor:
Updated, please check.

tiann:
`DAC_READ_SEARCH` is enough, don't use `DAC_OVERRIDE`

tiann:
I don't test if if can autostart when boot if we only give it `DAC_READ_SEARCH`

JohnRTitor:
Tested and updated. Looks like it can do with `DAC_READ_SEARCH` Thanks!

tiann (approved):

Issue #1331, opened 2024-02-01
[bug]5.4内核system分区挂载overlayfs失败


natsumerinchan:
### Please check before submitting an issue

- [X] I have searched the issues and haven't found anything relevant
- [X] I will upload bugreport file in KernelSU Manager - Settings - Report log
- [X] I know how to reproduce the issue which may not be specific to my device

### Describe the bug

自PR https://github.com/tiann/KernelSU/pull/1325 合并后我的设备上需要挂载system分区的模块会失效（如字体模块，nano-ndk等），但root功能正常

### To Reproduce

1.安装 PR https://github.com/tiann/KernelSU/pull/1325 以后的管理器
2.打开app
3.重启系统

### Logs

[KernelSU_bugreport_2024-02-01_21_56.tar.gz](https://github.com/tiann/KernelSU/files/14126616/KernelSU_bugreport_2024-02-01_21_56.tar.gz)

### Device info

- Device: Oneplus 9RT(martini)
- OS Version:Pixel Experience 13 (Plus edition) (Android13)
- KernelSU Version: kernelspace(11497) + userspace(11497)
- Kernel Version: 5.4.256

natsumerinchan:
在boot.log发现的错误
```
06-29 22:49:36.104   728   728 I KernelSU: ksud::module: load /data/adb/modules/Hc_memory system.prop
06-29 22:49:36.106   728   728 I KernelSU: ksud::module: load /data/adb/modules/yc_dfps system.prop
06-29 22:49:36.109   728   728 I KernelSU: ksud::event: module: /data/adb/modules/lost+found is disabled, ignore!
06-29 22:49:36.109   728   728 I KernelSU: ksud::event: module: /data/adb/modules/systemless-hosts-KernelSU-module skip_mount exist, skip!
06-29 22:49:36.110   728   728 I KernelSU: ksud::event: module: /data/adb/modules/ssh skip_mount exist, skip!
06-29 22:49:36.110   728   728 I KernelSU: ksud::mount: mount overlay for /system
[...]

PR #1410, opened 2024-03-02
ksud: Use fallback method to mount overlayfs when fsopen one failed

natsumerinchan:
Fix https://github.com/tiann/KernelSU/issues/1331 (Power by GPT-4.0)

### DIFF
--- a/userspace/ksud/src/utils.rs
+++ b/userspace/ksud/src/utils.rs
@@ -299,14 +299,6 @@ pub fn copy_module_files(source: impl AsRef<Path>, destination: impl AsRef<Path>
             log::info!("Symlink: {:?} -> {:?}", dest_path, target);
             std::os::unix::fs::symlink(target, &dest_path).context("Failed to create symlink")?;
             copy_xattrs(&source_path, &dest_path)?;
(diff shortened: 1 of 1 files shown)


tiann:
I think the method of judging versions is not good enough, because some people may cherry pick; it should be judged if the statement is successful, if it fails, go to else, instead of just judging the first fsopen.

natsumerinchan:
> I think the method of judging versions is not good enough, because some people may cherry pick; it should be judged if the statement is successful, if it fails, go to else, instead of just judging the first fsopen.

Done.

tiann:
Sorry, I made a mistake and accidentally closed this PR. I have re-submitted it at https://github.com/tiann/KernelSU/commit/3829894d4d205642bb241fc5846463729e4d3eba and added you as a co-author. Sorry again.

tiann:
I just don't why [this](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork) doesn't work :(

PR #1449, opened 2024-03-14
[CI][A13] a13-5.10 patch level 2024-02

apprehensions:
My Pixel 7a has a kernel version `5.10.177-android13-4-00003-ga7208022a7ea-ab1081582`, with a security patch of 2024-02, but the only available kernel build for that is security patch 2023-07 which causes a bootloop.


apprehensions:
Seems to have built 5.10.205-android13-4-g1a7524621846.

tiann (approved):

PR #1471, opened 2024-03-18
 [CI][A13] a13-5.10 patch level 2024-03

TheNeutrinoRaged:
I updated my pixel 7a to 5th March 2024 OTA. It has kernel version 5.10.189-android13-4-00012-g1217bb583cc5-ab11174560 and Android security update 2024-03, I can't see a 5.10 kernel in releases with 2024-03 security date. (edit: can't see 13-5.10*)

(I based this PR off https://github.com/tiann/KernelSU/pull/1449 if its wrong or missing info feel free to ignore)

Thanks


tiann:
sublevel is wrong: https://cs.android.com/android/kernel/superproject/+/common-android13-5.10-2024-03:common/Makefile

PR #1494, opened 2024-03-21
[RFC] kernel: try umount /system/etc/hosts

backslashxx:
Modding hosts file is a common thing to have on rooted android devices.
But this is easy to detect, either by reading the file, checking modified time and checking its filesize.
Unmounting it as needed for non-root processes solves this issue.

This is a PR requesting for comments for this change. 
Use it with this module https://github.com/backslashxx/bindhosts/


tiann:
I won't merge this, if someone wants to umount ca certs, do they also need to add it? You can mount it to /system, and then the kernel will give you an overall umount.

PR #1512, opened 2024-03-23
Update Workflows and Dependencies

CanerKaraca23:
Corrections would be very welcome.


.github/workflows/build-kernel-arcvm.yml
```diff
@@ -44,7 +44,7 @@ jobs:
 
           sudo ln -s --force python3 /usr/bin/python
 
-          export LLVM_VERSION=12
+          export LLVM_VERSION=18
```
tiann, line 47: You should not change the version of the compilation toolchain, whether it is AOSP or kernel source code, there are requirements for the compiler; AOSP even only allows you to use a fixed version of Ubuntu. No one is willing to fight with the compiler for no reason.

.github/workflows/wsa-kernel.yml
```diff
@@ -33,14 +33,14 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ./llvm
-          key: llvm-12.0.1
+          key: llvm-18.1.2
```
tiann, line 36: same here

.github/workflows/wsa-kernel.yml
```diff
@@ -33,14 +33,14 @@ jobs:
...
 
       - name: Setup LLVM
         uses: KyleMayes/install-llvm-action@v1
         with:
-          version: "12.0.1"
+          version: "18.1.2"
```
tiann, line 41: same here

tiann (changes requested):

tiann (approved):

PR #1543, opened 2024-03-27
kernel: fix issue with dfd parameter retrieval

18712886438:
Although in most cases, this system call might not be used, or when called, AT_FDCWD(-100) is passed in.


tiann (approved):

tiann:
`android12-5.10-2023-06` is removed by Google, just ignore it.

PR #1545, opened 2024-03-27
kernel:Add Huawei hisi check

Coconutat:
Use huawei_hisi_check.h to determine whether it is an old Huawei HiSilicon device.   
Solve:  
1. Compatible with non-GKI Huawei HiSilicon devices
2. Solve different bugs in EMUI of different system versions
3. Does not affect other devices


kernel/huawei_hisi_check.h
```diff
@@ -0,0 +1,22 @@
+#ifndef _HUAWEI_HISI_CHECK_H
+#define _HUAWEI_HISI_CHECK_H
+
+
+#include "linux/version.h"
+#include "ss/policydb.h"
+
+/*
+ * Adapt to Huawei HISI kernel without affecting other kernels ,
```
tiann, line 9: please put this in `kernel_compat.h`

tiann (changes requested):

kernel/huawei_hisi_check.h
Coconutat: OK , Fix it

tiann (approved):

PR #1549, opened 2024-03-28
kernel: no need to umount and mark as root for manager uid

5ec1cff:


kernel/allowlist.c
```diff
@@ -289,7 +295,12 @@ bool __ksu_is_allow_uid(uid_t uid)
 bool ksu_uid_should_umount(uid_t uid)
 {
 	struct app_profile profile = { .current_uid = uid };
-	bool found = ksu_get_app_profile(&profile);
+	bool found;
+	if (likely(ksu_is_manager_uid_valid()) && unlikely(ksu_get_manager_uid() == uid)) {
+		// we should not umount on manager!
+		return false;
+	}
+	found = ksu_get_app_profile(&profile);
```
yujincheng08, line 303:
```suggestion
	if (likely(ksu_is_manager_uid_valid()) && unlikely(ksu_get_manager_uid() == uid)) {
		// we should not umount on manager!
		return false;
	}
	bool found = ksu_get_app_profile(&profile);
```

PR #1586, opened 2024-04-03
throne_tracker: Fix Manager sometimes not detecting KSU and optimize it

bemerguy:
The original logic was wrong and used 3 strlen()s for every file found, wasting cpu.

Optimize it by first comparing only the filename length, given we already know it,
and then strncmp() to compare with "base.apk"

Tested successfully on my Bandido Kernel (4.19)


tiann (approved):

tiann:
thank you!

PR #1657, opened 2024-04-23
Hook syscalls and stable symbols

anotherjin:
1. Replace `do_execveat_common` with `sys_execve` and `sys_execveat`
2. Replace `input_handle_event` with `input_event` and `input_inject_event`

Tested on android12-5.10-2024-04, android13-5.15-2024-04. android14-6.1-2024-04

### DIFF
--- a/kernel/ksud.c
+++ b/kernel/ksud.c
@@ -263,119 +263,6 @@ int ksu_handle_execveat_ksud(int *fd, struct filename **filename_ptr,
 	return 0;
 }
 
-int ksu_handle_execve_ksud(int *fd, const char **filename_ptr,
-				struct user_arg_ptr *argv, struct user_arg_ptr *envp, int *flags)
-{
-	static const char app_process[] = "/system/bin/app_process";
-	static bool first_app_process = true;
-
-	/* This applies to versions Android 10+ */
-	static const char system_bin_init[] = "/system/bin/init";
-	/* This applies to versions between Android 6 ~ 9  */
-	static const char old_system_init[] = "/init";
-	static bool init_second_stage_executed = false;
-
-	char filename[sizeof(app_process) + 1];
-
-#ifndef CONFIG_KPROBES
-	if (!ksu_execveat_hook) {
-		return 0;
-	}
-#endif
-
-	if (!filename_ptr)
-		return 0;
-
-	memset(filename, 0, sizeof(filename));
-	ksu_strncpy_from_user_nofault(filename, *filename_ptr, sizeof(filename));
-
-	if (unlikely(!memcmp(filename, system_bin_init, 
-				sizeof(system_bin_init) - 1) && argv)) {
-		// /system/bin/init executed
-		int argc = count(*argv, MAX_ARG_STRINGS);
-		pr_info("/system/bin/init argc: %d\n", argc);
-		if (argc > 1 && !init_second_stage_executed) {
-			const char __user *p = get_user_arg_ptr(*argv, 1);
-			if (p && !IS_ERR(p)) {
-				char first_arg[16];
-				ksu_strncpy_from_user_nofault(first_arg, p, sizeof(first_arg));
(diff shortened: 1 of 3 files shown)


tiann:
There are generally several scenarios where `su` is needed:

1. Call `su` through shell, i.e. `/system/bin/sh`
2. Call `Runtime.exec()` directly in Java
3. Call execve directly in jni

Therefore, I think we don't need to handle so many execve calls, because:

1. The `compat_` series is not needed; because we don't support 32-bit ourselves; for the case of "kernel is 64-bit, but system is 32-bit", this situation did exist in the past, but I think it is unnecessary to be compatible. If you must be compatible, you can only deal with non-GKI.
2. The `execveat` series does not need to be handled; because
   - Android's shell `ash` always uses execve when executing commands (http://aospxref.com/android-14.0.0_r2/xref/external/mksh/src/exec.c)
   - `Runtime.exec` always uses execve (http://aospxref.com/android-14.0.0_r2/xref/libcore/ojluni/src/main/native/UNIXProcess_md.c#666)
   - If you write jni code yourself, you can use execve directly;

I think it should work fine to remove those three kprobes.

kernel/sucompat.c
```diff
@@ -196,6 +197,21 @@ static int execve_handler_pre(struct kprobe *p, struct pt_regs *regs)
...
+static int sys_execve_handler_pre(struct kprobe *p, struct pt_regs *regs)
+{
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 16, 0)
+	struct pt_regs *real_regs = (struct pt_regs *)PT_REGS_PARM1(regs);
+#else
+	struct pt_regs *real_regs = regs;
+#endif
+	const char __user *filename_user =
+		(const char *)PT_REGS_PARM1(real_regs);
+	struct filename *filename_mid = getname(filename_user);
```
tiann, line 209: `getname` may cause the CPU to sleep due to memory allocation, which is not allowed in kprobe handlers; this usage may lead to random kernel panic. You need to manually copy the memory using the `_nofault` series of functions and then directly compare the strings; reference: ksu_handle_faccessat.

kernel/ksud.c
```diff
@@ -614,6 +650,6 @@ void ksu_ksud_exit() {
 	unregister_kprobe(&execve_kp);
 	// this should be done before unregister vfs_read_kp
 	// unregister_kprobe(&vfs_read_kp);
-	unregister_kprobe(&input_handle_event_kp);
+	unregister_kprobe(&input_event_kp);
```
tiann, line 653: `unregister_kprobes(input_event_kps, 2);` ?

tiann (changes requested):

kernel/ksud.c
```diff
@@ -263,6 +263,119 @@ int ksu_handle_execveat_ksud(int *fd, struct filename **filename_ptr,
 	return 0;
 }
 
+int ksu_handle_execve_ksud(int *fd, const char **filename_ptr,
```
tiann, line 266: Please "Don't repeat yourself", try merge the logic with `ksu_handle_execveat_ksud`, they are many same logic, we don't want to maintain two copies.

tiann: If it works well below 5.10, just delete `ksu_handle_execveat_ksud` is OK

tiann (changes requested):

tiann (dismissed):

tiann:
> drivers/kernelsu/sucompat.c:278:17: error: use of undeclared identifier 'SYS_EXECVE_SYMBOL'
        .symbol_name = SYS_EXECVE_SYMBOL,
                       ^
1 error generated.
make[2]: *** [scripts/Makefile.build:287: drivers/kernelsu/sucompat.o] Error 1
make[1]: *** [scripts/Makefile.build:549: drivers/kernelsu] Error 2
make: *** [Makefile:1957: drivers] Error 2

x86_64 failed

tiann (approved):

PR #1681, opened 2024-04-29
[Experimental] Release magiskboot binaries to /data/adb

fsxitutu:
Experimental!
 Added the release of magiskboot to /data/adb during the process of ksu manager releasing ksud to /data/adb

### DIFF
--- a/manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
+++ b/manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
@@ -61,7 +61,7 @@ fun HomeScreen(navigator: DestinationsNavigator) {
         ) {
             val isManager = Natives.becomeManager(ksuApp.packageName)
             SideEffect {
-                if (isManager) install()
+                if (isManager) install(File(ksuApp.applicationInfo.nativeLibraryDir, "libmagiskboot.so"))
             }
             val ksuVersion = if (isManager) Natives.version else null
             val lkmMode = ksuVersion?.let {
--- a/manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
+++ b/manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
@@ -61,7 +61,7 @@ fun HomeScreen(navigator: DestinationsNavigator) {
         ) {
             val isManager = Natives.becomeManager(ksuApp.packageName)
             SideEffect {
-                if (isManager) install(File(ksuApp.applicationInfo.nativeLibraryDir, "libmagiskboot.so"))
+                if (isManager) install()
             }
             val ksuVersion = if (isManager) Natives.version else null
             val lkmMode = ksuVersion?.let {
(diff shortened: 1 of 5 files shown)


manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
```diff
@@ -61,7 +61,7 @@ fun HomeScreen(navigator: DestinationsNavigator) {
         ) {
             val isManager = Natives.becomeManager(ksuApp.packageName)
             SideEffect {
-                if (isManager) install()
+                if (isManager) install(File(ksuApp.applicationInfo.nativeLibraryDir, "libmagiskboot.so"))
```
tiann, line 64: don't need to pass parameter here.

manager/app/src/main/java/me/weishu/kernelsu/ui/util/KsuCli.kt
```diff
@@ -82,9 +82,9 @@ fun execKsud(args: String, newShell: Boolean = false): Boolean {
     }
 }
 
-fun install() {
+fun install(magiskboot_path: String) {
     val start = SystemClock.elapsedRealtime()
-    val result = execKsud("install", true)
+    val result = execKsud("install --magiskboot $magiskboot_path", true)
```
tiann, line 87: concat the `magiskboot` path here is ok

userspace/ksud/src/defs.rs
```diff
@@ -12,9 +12,11 @@ pub const PROFILE_TEMPLATE_DIR: &str = concatcp!(PROFILE_DIR, "templates/");
 pub const KSURC_PATH: &str = concatcp!(WORKING_DIR, ".ksurc");
 pub const KSU_OVERLAY_SOURCE: &str = "KSU";
 pub const DAEMON_PATH: &str = concatcp!(ADB_DIR, "ksud");
+pub const MAGISKBOOT_PATH: &str = concatcp!(ADB_DIR, "magisk");
```
tiann, line 15: We can put `magiskboot` directly in `/data/adb/ksu/bin/`. There is no need to create a symlink, as `/data/adb/ksu/bin` is already added to the `$PATH` for su. You can use `magiskboot` without specifying its full path.

userspace/ksud/src/utils.rs
```diff
@@ -1,393 +1,400 @@
...
-        }
-    }
-    Ok(())
-}
-
-#[cfg(not(any(target_os = "linux", target_os = "android")))]
-pub fn copy_module_files(_source: impl AsRef<Path>, _destination: impl AsRef<Path>) -> Result<()> {
-    unimplemented!()
-}
+use anyhow::{bail, Context, Error, Ok, Result};
```
tiann, line 1: the format seems incorrect, can you run `cargo format` first?

tiann (changes requested):

userspace/ksud/src/utils.rs
fsxitutu: I have not made any changes to the copy_module_files function, which comes from the original userspace/ksud/src/utils.rs file.  The whole file is marked red seems to be a bug

fsxitutu: In order to solve the difference comparison problem, I reopened a new PR, please refer to the new PR.

PR #1684, opened 2024-04-30
Add magiskboot to kernelsu working directory

fsxitutu:
Add the magiskboot binary to the /data/adb/ksu/bin directory so that scripts/programs can call magiskboot to patch the boot/init_boot image.


userspace/ksud/src/cli.rs
```diff
@@ -307,7 +307,7 @@ pub fn run() -> Result<()> {
                 Module::Shrink => module::shrink_ksu_images(),
             }
         }
-        Commands::Install => utils::install(),
+        Commands::Install { magiskboot } => utils::install(magiskboot),
```
tiann, line 310: You should add cli parameter here: https://github.com/tiann/KernelSU/pull/1684/files#diff-50c7b7af6d0fd53a343c78c5b80654b053d158d7e3b24e744040165af53cc505R38

userspace/ksud/src/utils.rs
```diff
@@ -208,10 +208,12 @@ fn link_ksud_to_bin() -> Result<()> {
     Ok(())
 }
 
-pub fn install() -> Result<()> {
+pub fn install(magiskboot: Option<PathBuf>) -> Result<()> {
     ensure_dir_exists(defs::ADB_DIR)?;
     std::fs::copy("/proc/self/exe", defs::DAEMON_PATH)?;
+    std::fs::copy(magiskboot, defs::MAGISKBOOT_PATH)?;
```
tiann, line 214: Please add the following two lines to the end of the function. This should not affect the existing logic if it fails.

tiann (changes requested):

tiann:
please fix the two issues, others look good to me

userspace/ksud/src/utils.rs
```diff
@@ -219,6 +219,9 @@ pub fn install() -> Result<()> {
     link_ksud_to_bin()?;
 
     Ok(())
```
tiann, line 221: This should be the last one

tiann (changes requested):

userspace/ksud/src/utils.rs
fsxitutu: Capture possible errors to avoid affecting normal operation。utils.rs Already uploaded

tiann (approved):

Issue #270, opened 2023-02-23
移植到3.18内核后(不是kprobe模式)，root工作，modules不工作


kodhh:
dmesg | grep "ksu"  如下
[    7.076276] init: starting service 'exec 13 (/data/adb/ksud services)'...
[    7.079127] init: SVC_EXEC service 'exec 13 (/data/adb/ksud services)' pid 1059 (uid 0 gid 0+0 context u:r:su:s0) started; waiting...
[    7.097453] init: cannot execve('/data/adb/ksud'): Permission denied
[    7.101245] init: Service 'exec 13 (/data/adb/ksud services)' (pid 1059) exited with status 127 waiting took 0.022000 seconds
[   20.850326] init: starting service 'exec 15 (/data/adb/ksud boot-completed)'...
[   20.852763] init: SVC_EXEC service 'exec 15 (/data/adb/ksud boot-completed)' pid 2763 (uid 0 gid 0+0 context u:r:su:s0) started; waiting...
[   20.852900] init: cannot execve('/data/adb/ksud'): Permission denied
[   20.854885] init: Service 'exec 15 (/data/adb/ksud boot-completed)' (pid 2763) exited with status 127 waiting took 0.002000 seconds

请问大佬，这是怎么了？

tiann:
```
cannot execve('/data/adb/ksud'): Permission denied
```

> normally ksud will be started by the init process (initrc is injected to it by ksu from kernel) on boot and it handles module updates, mountings and post-fs-data or service scripts execution

> ksud requires some privileges to run ofc, so it must change its selinux domain from init to su(ksu domain)

> but ksud is located under /data which has the nosuid mount flag, domain transitions are restricted by selinux in such environment (only bounded transitions are allowed but it does not meet our needs), this results in the ksud not starting at all on boot and therefore the modules cannot work

> bcs we didn't find a proper way to fix unbounded domain transitions under nnp/nosuid environment on kernels earlier than 4.14, u should patch ksu or kernel for now to fix the module support

tiann:
refer this: https://github.com/ookiineko/KernelSU/commit/0950fbba5e79820da539054f6eb6d1d48ec96ff4

powellnorma:
> refer this: https://github.com/ookiineko/KernelSU/commit/0950fbba5e79820da539054f6eb6d1d48ec96ff4

Link does not work anymore. But the commit is also accessible here:
[...]

PR #1704, opened 2024-05-07
kernel: selinux: add security_bounded_transition hook for kernel < 4.14

powellnorma:
- https://github.com/torvalds/linux/commit/af63f4193f9fbbbac50fc766417d74735afd87ef

- SELinux domain transitions under NNP/nosuid environment was introduced in 4.14 by the above commit, for older kernels, we need to make sure our domain transitions are allowed when calling ksud at boot from the init

- Adapted from https://github.com/tiann/KernelSU/issues/270#issuecomment-2097706460 https://github.com/ookiineko/KernelSU/commit/0950fbba5e79820da539054f6eb6d1d48ec96ff4

### DIFF
--- a/kernel/Makefile
+++ b/kernel/Makefile
@@ -11,6 +11,7 @@ kernelsu-objs += kernel_compat.o
 kernelsu-objs += selinux/selinux.o
 kernelsu-objs += selinux/sepolicy.o
 kernelsu-objs += selinux/rules.o
+kernelsu-objs += selinux/kernel_compat.o
 ccflags-y += -I$(srctree)/security/selinux -I$(srctree)/security/selinux/include
 ccflags-y += -I$(objtree)/security/selinux -include $(srctree)/include/uapi/asm-generic/errno.h
 
--- a/kernel/ksu.c
+++ b/kernel/ksu.c
@@ -37,6 +37,10 @@ extern void ksu_sucompat_exit();
 extern void ksu_ksud_init();
 extern void ksu_ksud_exit();
 
+#if LINUX_VERSION_CODE < KERNEL_VERSION(4, 14, 0)
+extern void ksu_enable_selinux_compat();
+#endif
+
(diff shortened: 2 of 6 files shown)


tiann:
Kprobe handlers cannot perform operations that abandon the CPU, and many of the functions here do not meet the conditions, which may cause a kernel panic.

powellnorma:
> Kprobe handlers cannot perform operations that abandon the CPU

What do you with "abondon"? I think the code just checks if the current transition is a particular one (init -> su), and if so, changes the value of `old_sid`.

Also I think the kprobe just places a jmp instructions to the hook - So unless the function is called in a special context, I think the hook can do whatever it wants. Am I wrong?

Or does `security_bounded_transition` get called in such a "special context"?

tiann:
kprobe itself is a special context, you cannot call code that might sleep

powellnorma ***@***.***> 于 2024年5月7日周二 下午7:38写道：

tiann:
1. LKM is ok because it is just some predefined callback.
2. this is possible, it is kernel inline hook actually, there are some projects on github already, but it is not stable than kprobe.

We may add inline hook in the future if we have time.

tiann:
`security_secctx_to_secid` has lock stuff.

powellnorma:
> security_secctx_to_secid has lock stuff.

Sorry for my ignorance, but how did you find out? Have you tested it, and it led to crashes?

tiann:
> > security_secctx_to_secid has lock stuff.
> 
> Sorry for my ignorance, but how did you find out? Have you tested it, and it led to crashes?

I just checked the source code.

powellnorma:
> I just checked the source code.

But where does it use locks? Or am I looking at the wrong spot?

https://elixir.bootlin.com/linux/v4.4.3/source/security/security.c#L1151

Maybe the actual selinux implementation uses locks - But the kprobe is placed before that is invoked, IIUC

PR #1713, opened 2024-05-08
ksud: refine get_tmp_path

### LINKED ISSUE #1710: ADB not working after flashing a simple module

- [X] I have searched the issues and haven't found anything relevant
- [X] I will upload bugreport file in KernelSU Manager - Settings - Report log
- [ ] I know how to reproduce the issue which may not be specific to my device

### Describe the bug

After flashing a simple module that just overwrites `/system/etc/hosts` and rebooting, adb is not working anymore. The device simply does not show up anymore, no matter what I try.

Deleting `/data/adb/ksu/modules.img` via termux and rebooting fixes the issue.
In termux I can also see that `/system/etc/hosts` does get correctly replaced / over-mounted.

One thing that looks suspicious to me:

```
[   18.812591] [pid:1,cpu0,init]init: Service 'exec 5 (/data/adb/ksud)' (pid 701) exited with status 0 waiting took 0.150053 seconds
[   18.812652] [pid:1,cpu0,init]init: Wait for exec took 0.150207 seconds
[   18.814117] [pid:1,cpu0,init]init: processing action (sys.usb.config=hisuite,mtp,mass_storage,adb && sys.usb.configfs=1)
[   18.814331] [pid:1,cpu0,init]init: write_file: Unable to write to '/config/usb_gadget/g1/UDC': No such device
[   18.814727] [pid:1,cpu0,init]init: Service 'macaddr' (pid 693) exited with status 0
[   18.815124] [pid:1,cpu0,init]init: cannot find '/sbin/adbd', disabling 'adbd': No such file or directory
```

In dmesg.log.old, `cannot find '/sbin/adbd', disabling 'adbd'` does not appear.

I know this kernel is quite old, and it could be a weird kernel bug. So I don't expect anyone to spend much time with this. Just wanted to ask in case someone as an idea. Thank you!

[KernelSU_bugreport_2024-05-08_10_23.tar.gz](https://github.com/tiann/KernelSU/files/15245969/KernelSU_bugreport_2024-05-08_10_23.tar.gz)

### To Reproduce

_No response_

### Expected behavior

_No response_

### Screenshots

_No response_

### Logs

_No response_

### Device info

- Device: HUAWEI P8 lite 2017
- OS Version: Android 8
- KernelSU Version: 11823
- Kernel Version: 4.4.23 (am using https://github.com/riarumoda/KernelSU-4.4)

### Additional context

_No response_

powellnorma:
Fixes #1710

What do you think?

* It first uses `TEMP_DIR` (`/debug_ramdisk`) if it exists + is is empty.
* Otherwise it tries to create a random directory in `/dev`.
* If that fails, it goes through a list of directories (including `TEMP_DIR_LEGACY`), and chooses the first one that is empty.
* If no empty directory it chooses the first one that exists

### DIFF
--- a/userspace/ksud/src/utils.rs
+++ b/userspace/ksud/src/utils.rs
@@ -1,12 +1,9 @@
 use anyhow::{bail, Context, Error, Ok, Result};
 use std::{
-    fs::{self, create_dir_all, remove_file, write, File, OpenOptions},
-    io::{
+    fs::{self, create_dir_all, remove_file, write, File, OpenOptions}, io::{
         ErrorKind::{AlreadyExists, NotFound},
         Write,
-    },
-    path::Path,
-    process::Command,
+    }, path::Path, process::Command, sync::OnceLock
 };
 
 use crate::{assets, boot_patch, defs, ksucalls, module, restorecon};
@@ -199,16 +196,21 @@ fn is_ok_empty(dir: &str) -> bool {
 
 fn find_temp_path() -> String {
     use std::result::Result::{Ok, Err};
-    use tempfile::Builder;
 
     if is_ok_empty(defs::TEMP_DIR) {
         return defs::TEMP_DIR.to_string();
     }
 
     // Try to create a random directory in /dev/
-    let r = Builder::new().tempdir_in("/dev/");
+    let r = tempdir::TempDir::new_in("/dev/", "");
     match r {
-        Ok(tmp_dir) => {return tmp_dir.path().to_owned().to_str().unwrap().to_string()},
+        Ok(tmp_dir) => {
+            let buf = tmp_dir.path().to_owned();
+            let s = buf.to_str();
+            if s.is_some() {
+                return s.unwrap().to_string()
+            }
+        },
         Err(_e) => {}
     }
 
@@ -222,14 +224,14 @@ fn find_temp_path() -> String {
 
     // find empty directory
     for dir in dirs {
-        if is_ok_empty(defs::TEMP_DIR) {
+        if is_ok_empty(dir) {
             return dir.to_string();
         }
     }
 
     // Fallback to non-empty directory
     for dir in dirs {
-        if metadata(defs::TEMP_DIR).is_ok() {
+        if metadata(dir).is_ok() {
             return dir.to_string();
         }
     }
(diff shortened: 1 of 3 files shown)


userspace/ksud/src/utils.rs
```diff
@@ -188,14 +188,71 @@ pub fn has_magisk() -> bool {
...
+        defs::TEMP_DIR,
+        "/patch_hw",
+        "/oem",
+        "/root",
+        defs::TEMP_DIR_LEGACY,
+    ];
+
+    // find empty directory
+    for dir in dirs {
+        if is_ok_empty(defs::TEMP_DIR) {
```
tiann, line 225:
```suggestion
        if is_ok_empty(dir) {
```

userspace/ksud/src/utils.rs
```diff
@@ -188,14 +188,71 @@ pub fn has_magisk() -> bool {
...
+fn find_temp_path() -> String {
+    use std::result::Result::{Ok, Err};
+    use tempfile::Builder;
+
+    if is_ok_empty(defs::TEMP_DIR) {
+        return defs::TEMP_DIR.to_string();
+    }
+
+    // Try to create a random directory in /dev/
+    let r = Builder::new().tempdir_in("/dev/");
```
tiann, line 209: Can the `tempdir` crate archive this?

userspace/ksud/src/utils.rs
```diff
@@ -188,14 +188,71 @@ pub fn has_magisk() -> bool {
...
+    match r {
+        Ok(tmp_dir) => {return tmp_dir.path().to_owned().to_str().unwrap().to_string()},
```
tiann, line 211: please don't use unwrap here

userspace/ksud/src/utils.rs
```diff
@@ -188,14 +188,71 @@ pub fn has_magisk() -> bool {
...
+            return dir.to_string();
+        }
+    }
+
+    "".to_string()
+}
+
+
+pub fn get_tmp_path() -> String {
+    static mut CHOSEN_TMP_PATH: Option<String> = None;
```
tiann, line 242: `static mut` is not good, please use `once_cell`

userspace/ksud/src/utils.rs
```diff
@@ -188,14 +188,71 @@ pub fn has_magisk() -> bool {
...
+    let r = Builder::new().tempdir_in("/dev/");
+    match r {
+        Ok(tmp_dir) => {return tmp_dir.path().to_owned().to_str().unwrap().to_string()},
+        Err(_e) => {}
+    }
+
+    let dirs = [
+        defs::TEMP_DIR,
+        "/patch_hw",
+        "/oem",
```
tiann, line 218: is `/oem` and `/root` safe to use?

tiann (changes requested):

userspace/ksud/src/utils.rs
powellnorma: Probably not for every ROM - that's why they are used only as fallback, and even then only when they are empty (or no empty directory was found and previous options did not exist). `/sbin` arguably is not (completely) safe as well (https://github.com/tiann/KernelSU/issues/1710).

We could switch the order in the `dirs` list however - the "safest" options should come first

tiann (approved):

tiann (approved):

PR #1733, opened 2024-05-15
Refactor checkNewVersion function to use LatestVersionInfo data class

ALEX5402:
- Updated the checkNewVersion function to return a LatestVersionInfo data class instead of a Triple.
- Defined default null value for LatestVersionInfo in case of failure.
- Improved readability and maintainability by replacing the Triple with a data class.
- Included version code, download URL, and changelog in the LatestVersionInfo data class.

### DIFF
--- a/manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
+++ b/manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
@@ -38,6 +38,7 @@ import me.weishu.kernelsu.ui.component.rememberConfirmDialog
 import me.weishu.kernelsu.ui.screen.destinations.InstallScreenDestination
 import me.weishu.kernelsu.ui.screen.destinations.SettingScreenDestination
 import me.weishu.kernelsu.ui.util.*
+import me.weishu.kernelsu.ui.util.module.LatestVersionInfo
 
 @RootNavGraph(start = true)
 @Destination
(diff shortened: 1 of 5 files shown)


manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
```diff
@@ -100,13 +101,22 @@ fun HomeScreen(navigator: DestinationsNavigator) {
 @Composable
 fun UpdateCard() {
     val context = LocalContext.current
-    val newVersion by produceState(initialValue = Triple(0, "", "")) {
-        value = withContext(Dispatchers.IO) { checkNewVersion() }
+    val latestVersionInfo = LatestVersionInfo(
```
tiann, line 104: we can use default constructor

tiann (changes requested):

manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
ALEX5402: yaa but using a data class should be a better option because data class gives more flexibility i should recommend it

tiann: data class can have default constructor, you don't need to pass it everytime

tiann: you can modify this PR directly, i will squash it when merge.

Issue #1664, opened 2024-04-25
[Feature] Add logs export to local storage


Fede2782:
### Is your feature request related to a problem? Please describe.

I'm always frustrated when I need to get KernelSU logs because when I click the Logs button in Settings it opens the Share menu, this means that if I have to get the logs file on the device it has been captured on, I need to share it to another device first or upload to some cloud service. 

### Describe the solution you'd like.

It should open a popup which asks if the user wants to share or save on device. In case the user wants to save on device a DocumentsUI screen or similar should allow the user to save the logs where he desires on by default in some Download folders

### Describe alternatives you've considered.

Instead of sharing it and then deleting it at least give the user the path where the log is saved

powellnorma:
Workarounds:

https://f-droid.org/en/packages/xyz.myachin.saveto/

https://android.stackexchange.com/questions/195062/share-as-file-on-filesystem

LHC12580:
到底是什么情况啊？



---Original---
From: ***@***.***&gt;
Date: Tue, May 7, 2024 20:41 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [tiann/KernelSU] [Feature] Add logs export to local storage(Issue #1664)




  
韦树哥为什么不更新了，腾讯和阿里又威胁你们了是吗？
  
又是哪个自媒体在那乱说不更新了，能不能先提升自己的英语水平，把公告看懂；再不济，Google 翻译都不会用吗？

klxiaoniu:
How to design this interaction? Simply replacing sending logs with save logs, like LSPosed, or add a dialog that allows to choose save or share?

tiann:
I think save log is OK, if he wants to share, he can share it in file
manager later.

快乐小牛 ***@***.***> 于 2024年5月9日周四 下午4:37写道：

Fede2782:
> Or we can add an Intent to ourselves, and when we share to this intent, it will be saved locally.

I think the best way could be to click on Capture logs -> Let the user chose if he wants to save locally or wants to share. In the first case a DocumentsUI page could be opened to let user save where he wants. In the second case the share menu could open

tiann:
#1741

yousefboutrad:
Kernel samsonj m 53 !!

nasryhsyb00-jpg:
> #3629 )

PR #1741, opened 2024-05-18
Added a feature to save logs to local storage

ALEX5402:
- Added export logs to local storage [1664](https://github.com/tiann/KernelSU/issues/1664)


manager/app/build.gradle.kts
```diff
@@ -23,8 +23,8 @@ android {
 
     buildTypes {
         release {
-            isMinifyEnabled = true
-            isShrinkResources = true
+            isMinifyEnabled = false
```
tiann, line 26: please don't touch this

manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Settings.kt
```diff
@@ -151,6 +158,49 @@ fun SettingScreen(navigator: DestinationsNavigator) {
                 enableWebDebugging = it
             }
 
+            ListItem(
+                leadingContent = {
+                    Icon(
+                        Icons.Filled.Save,
+                        stringResource(id = R.string.save_log)
```
tiann, line 165: We already have "send log". It is better to merge these two.

tiann (changes requested):

manager/app/build.gradle.kts
ALEX5402: ok let me fix it i forgot to change this after test

manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Settings.kt
ALEX5402: ok got it

ALEX5402: But having this feature inside this app makes it more convenient and then user don't need to install any other apps just to do this simple thing

tiann:
we don't need two entries in settings, "save log" and "send log", it's strange, we only need one entry, when clicked, you can prompt user to choose

ALEX5402:
> we don't need two entries in settings, "save log" and "send log", it's strange, we only need one entry, when clicked, you can prompt user to choose

ok got it 

> choose

can i add a alert dialog there to choose user what to do ?

tiann:
> can i add a alert dialog there to choose user what to do ?

that's ok

ALEX5402:
> > can i add a alert dialog there to choose user what to do ?
> 
> that's ok

done sir please check this [commit](https://github.com/ALEX5402/KernelSU/commit/220ef324d0594bb3d52db8d0903fec2e39b9014d)

tiann (approved):

PR #1746, opened 2024-05-19
ksud: turn off auto-detect bootdevice when OS is not android

chise0713:


tiann:
This is too complicated, create a compile-time conditional variable, set it to true when target_os is android; if false, return directly, this implementation is simpler.

tiann (approved):

PR #1769, opened 2024-05-26
Rename what-is-kernelsu.md to what-is-kernelsuandroid13-5.10.190_2023…

kvqw112233:
…-12-boot-gz.img.gz.md


Fede2782:
Why? This is not an actual change either

Fede2782 (changes requested):
Unuseful

Issue #1771, opened 2024-05-26
throne_tracker.c building error


Pr0pHesyer:
### Please check before submitting an issue

- [X] I have searched the issues and haven't found anything relevant
- [X] I will upload bugreport file in KernelSU Manager - Settings - Report log
- [X] I know how to reproduce the issue which may not be specific to my device

### Describe the bug

Unofficial kernel building error.
Used kernel:https://github.com/LineageOS/android_kernel_xiaomi_msm8998
Building error:
/home/runner/work/Android-Builder/Android-Builder/Mi6/drivers/kernelsu/throne_tracker.c:173:54: error: too many arguments to function call, expected 2, have 3
                        unsigned int hash = full_name_hash(NULL, dirpath, strlen(dirpath));
                                            ~~~~~~~~~~~~~~                ^~~~~~~~~~~~~~~
/home/runner/work/Android-Builder/Android-Builder/Mi6/include/linux/dcache.h:89:21: note: 'full_name_hash' declared here
extern unsigned int full_name_hash(const unsigned char *, unsigned int);
                    ^
1 error generated.
make[3]: *** [/home/runner/work/Android-Builder/Android-Builder/Mi6/scripts/Makefile.build:292: drivers/kernelsu/throne_tracker.o] Error 1
make[2]: *** [/home/runner/work/Android-Builder/Android-Builder/Mi6/scripts/Makefile.build:527: drivers/kernelsu] Error 2
make[1]: *** [/home/runner/work/Android-Builder/Android-Builder/Mi6/Makefile:1080: drivers] Error 2

Detail in https://github.com/Pr0pHesyer/Android-Builder/actions/runs/9243917807/job/25428620808

### To Reproduce

1. clone kernel source
2. get toolchains
3. start build

### Expected behavior

Finish and output without error.

### Screenshots

_No response_

### Logs

```
/home/runner/work/Android-Builder/Android-Builder/Mi6/drivers/kernelsu/throne_tracker.c:173:54: error: too many arguments to function call, expected 2, have 3
                        unsigned int hash = full_name_hash(NULL, dirpath, strlen(dirpath));
                                            ~~~~~~~~~~~~~~                ^~~~~~~~~~~~~~~
/home/runner/work/Android-Builder/Android-Builder/Mi6/include/linux/dcache.h:89:21: note: 'full_name_hash' declared here
extern unsigned int full_name_hash(const unsigned char *, unsigned int);
                    ^
1 error generated.
[...]

PR #1773, opened 2024-05-26
kernel: fix throne_tracker uncompile-able on 4.4 kernel (issue #1771)

rsuntk:
in kernel v4.7.10:
extern unsigned int __pure full_name_hash(const char *, unsigned int);

in kernel v4.8.0:
extern unsigned int __pure full_name_hash(const void *salt, const char *, unsigned int);


Pr0pHesyer:
It could build.
Didn't really test in phone.

rsuntk:
What do you mean? So, it did build? But you didn't test it yet or something?

Pr0pHesyer:
It can be build without errors. 
The build products are not tested on the device.

tiann (approved):

PR #1785, opened 2024-05-31
kernelsu: LSM: Identify modules by more than name

hmtheboy154:
This is based on commit "LSM: Identify modules by more than name" https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v6.8&id=f3b8788cde61b02f1e6c202f8fac4360e6adbafc

Which is a requirement to build on kernel 6.8+. I use 912 as the ID for KernelSU as that's the date the Github repo was created.


hmtheboy154:
image
uh oh

hmtheboy154:
Oh no I didn't think about this 🥲
https://elixir.bootlin.com/linux/v6.8/source/security/security.c#L558

hmtheboy154:
Extra patches might be needed, but this is directly into the kernel
Basically add `CONFIG_KSU`  into `LSM_CONFIG_COUNT` define. I think we might have to update the wiki for this @tiann

hmtheboy154:
image
yep, it's working now

hmtheboy154:
https://github.com/android-generic/kernel-zenith/commit/40384accf1913ad6ebbe8cb8125efa772ee3271b

tiann:
This won't work for LKM mode, we need another solution.

hmtheboy154:
> This won't work for LKM mode, we need another solution.

Find a way to reduce lsm active count to make it lower than `LSM_CONFIG_COUNT` and then load the LKM ?

hmtheboy154:
photo_2024-06-01_19-39-34

ok this is bad 🤔

hmtheboy154:
@tiann  seems like there are some changes so extra edit in kernel is needed
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/security/security.c?h=v6.12.2&id=417c5643cd67a55f424b203b492082035d0236c3
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/security/security.c?h=v6.12.2&id=d6bd12e80bf94b055def6ff708e76f836b4b17ad
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/include/linux/lsm_count.h?h=v6.12.2&id=d51e783c17bab0c139bf78d6bd9d1f66673f7903

Although they mentioned `static lsm calls`, is that something new that could be helpful for us ?

backslashxx:
> This won't work for LKM mode, we need another solution.

for 6.12 LKM you might really have to hook symbols again
GKI is fixable with either +1 on LSM as hmtheboy154 did, or manually hooking security hooks.

hmtheboy154:
With the fact that some recent phones start shipped with kernel 6.12, if anyone try this PR please remember that you need to put this into your kernel
https://github.com/android-generic/kernel-zenith/commit/7042991a5c9ddae256f3864a598017310bb5ff2c

or else you won't be able to boot

backslashxx:
@hmtheboy154 lsm hook hack is kill damn, ksu has to hook prctl and setuid on kprobe for LKM

hmtheboy154:
> @hmtheboy154 lsm hook hack is kill damn, ksu has to hook prctl and setuid on kprobe for LKM

if it can, I don't even know if LKM is possible after 6.8+

backslashxx:
its still doable, the issue is that they have to move the hooks to kprobe (prctl, setuid)
not great when theres actual apps that probe prctl nowadays, its just gonna make it easier for them

for gki, even this can be skipped, we can just manually hook on security/security.c
straight up like how 3.x integration, but yeah way dirtier code

PR #1808, opened 2024-06-05
guide:Fix Simplified Chinese Translation missing and add SU works but module does not work patch

Coconutat:
Fix Simplified Chinese Translation missing and add SU works but module does not work patch.


tiann:
Since non gki is dropped, we won't merge this.

Coconutat:
> Since non gki is dropped, we won't merge this.

Sure

PR #1814, opened 2024-06-10
manager: improve grammar in english

igormiguell:
Improved the grammar and fixed some typos. Feel free to let me know if there are any mistakes.

A question. Why is the Profile App not translatable? Would it be possible to make it translatable?


Fede2782 (changes requested):
Weblate must be used for translations, at least for the ones which don't require changing blocked words

tiann (approved):

PR #1896, opened 2024-07-20
remove /data/system/users/0/package-restrictions.xml in safe mode

HDTianRu:
sometimes, phone will bacome brick due to some system packages are frozen in mistakes
so remove /data/system/users/0/package-restrictions.xml to rescue phone in safe mode


Ylarod:
This shouldn't be ksu's work, I think

PR #1920, opened 2024-07-26
Update cross and NDK

CanerKaraca23:


yujincheng08:
- for ndk, we should wait for actions/runner-images#10343 and use the bundled one
- for cross, we should wait for the next release (using mainline may suddenly break CI)

Sec32fun32 (approved):

PR #1934, opened 2024-08-01
manager: Add flash script into KernelSU Manager

SakuraKyuo:
Add flash script and libbusybox.so into KernelSU Manager, with that, users can more easier to get KernelSU like how they get Magisk.

Test: Build Manager - Flash in Third-Party Recovery - Reboot - Check KernelSU

Result: It works well on GKI Devices(Tested on Oneplus Ace 3 Pro)

Needs to improve:
- Not sure some Xiaomi Devices will works well or not, waiting for test.
- Need to add Non-GKI devices detecte


SakuraKyuo:
Force-pushed [59483bc](https://github.com/tiann/KernelSU/commit/59483bc6bab24e794f2182c1e9f820ad899f2ccd) is used to delete some unused codes

PR #1935, opened 2024-08-02
manager: Add flash script into KernelSU Manager

SakuraKyuo:
Add flash script and libbusybox.so into KernelSU Manager, with that, users can more easier to get KernelSU like how they get Magisk.

Test: Build Manager - Flash in Third-Party Recovery - Reboot - Check KernelSU

Result: It works well on GKI Devices(Tested on Oneplus Ace 3 Pro)

Needs to improve:
- Not sure some Xiaomi Devices will works well or not, waiting for test.
- Need to add Non-GKI devices detecte


SakuraKyuo:
In commit https://github.com/tiann/KernelSU/pull/1935/commits/3306c462e80efae1e0498be7a22c060d5b856f86, we have already support flash KernelSU in AOSP Based Recovery

After Android 11, the AOSP-based Recovery no longer has /sbin, while TWRP and LineageOS-based Recovery retain the /sbin/sh interpreter to provide compatibility with this shell flashing script. TWRP also adds bash to /sbin.

In order to adapt to AOSP-Based Recovery, we canceled the use of /sbin/bash to run the installation script, and instead used KernelSU's built-in busybox to run the installation script to achieve the purpose of adapting to all recovery.

We don't need updater-script anymore, so keep it empty. The core part is in update-binary and addon folder.

tiann:
Sorry, we have no plans to support flashing KernelSU in Recovery; most new machines don't have third-party recovery support, and supporting old devices is a pain.
And, we have provided AnyKernel.zip, if you really want to do this, it is better to modify AnyKernel.zip to support flashing in LKM mode.

SakuraKyuo:
> Sorry, we have no plans to support flashing KernelSU in Recovery; most new machines don't have third-party recovery support, and supporting old devices is a pain. And, we have provided AnyKernel.zip, if you really want to do this, it is better to modify AnyKernel.zip to support flashing in LKM mode.

Thanks for your review and reply.

PR #1960, opened 2024-08-08
js: add getModuleId method

jeffpeng3:
Add a method to get module ID in JS.
resolves https://github.com/tiann/KernelSU/issues/1571

test module's index.html
image

The test module:
[moduleId_test.zip](https://github.com/user-attachments/files/16552451/moduleId_test.zip)

test module's result:
image

### DIFF
--- a/js/README.md
+++ b/js/README.md
@@ -109,3 +109,12 @@ Show a toast message.
 import { toast } from 'kernelsu';
 toast('Hello, world!');
 ```
+
+### getModuleId
+
+Get Module Id.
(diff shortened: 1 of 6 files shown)


Howard20181:
Useless, issue wants to return to the module installation directory instead of the ID
No developer doesn't know the ID of the current module.

jeffpeng3:
Looks like I misunderstood the issue, apparently the full path is needed and not just the module ID, I've opened a new PR to implement the API.

PR #1981, opened 2024-08-15
Drop some legacy codes

yujincheng08:


Ylarod (approved):

yujincheng08:
I dont think it affects manager.

yujincheng08:
strange. these are only for mount. installing modules wont need mount.

yujincheng08:
Oh forgot that installing modules requires mounting. But anyway it should be okay; it's time to drop more support for non-gki.

chiteroman:
I can boot my 4.14 kernel without reverting this, is it really necessary to revert this commit?

CanerKaraca23:
> I can boot my 4.14 kernel without reverting this, is it really necessary to revert this commit?

You can't install modules with this commit on non-gki.

PR #2007, opened 2024-08-24
Update translation in pt-br and en-us

igormiguell:


tiann:
please use https://hosted.weblate.org/projects/kernelsu/manager/ thank you.

PR #2057, opened 2024-09-11
js: add moduleInfo method

jeffpeng3:
Add a `ksu.moduleInfo()` in JS.
resolves https://github.com/tiann/KernelSU/issues/1571

test module's index.html
image

The test module:
[moduleInfo_test.zip](https://github.com/user-attachments/files/17001977/moduleInfo_test.zip)

test module's result:
image

### DIFF
--- a/js/README.md
+++ b/js/README.md
@@ -109,3 +109,12 @@ Show a toast message.
 import { toast } from 'kernelsu';
 toast('Hello, world!');
 ```
+
+### moduleInfo
+
+Get Module info.
+```javascript
+import { moduleInfo } from 'kernelsu';
+// print moduleId in console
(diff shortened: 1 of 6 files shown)


jeffpeng3:
I'd like to know what you think should be there, such as whether the version needs to be provided, etc.

tiann:
You can refer the `ksud module list`

jeffpeng3:
I referenced `me.weishu.kernelsu.ui.util.listModules` and filtered out the current module's data, merged in the moduleDir and output it as a json string.

This should work the way you said and rely on the output of the `ksud module list`.

tiann (approved):

PR #2084, opened 2024-09-21
kernel: Fix compatibility with old and 32bit programs

### LINKED ISSUE #2045: kernelSU内核在0.9.3之后无法正确授权按键精灵root权限

- [X] I have searched the issues and haven't found anything relevant
- [X] I will upload bugreport file in KernelSU Manager - Settings - Report log
- [X] I know how to reproduce the issue which may not be specific to my device

### Describe the bug

kernelSU内核版本在0.9.3之后(也就是0.9.4到1.0.1)均无法正确授权按键精灵root权限，而在0.9.3及以下没有这个bug。

### To Reproduce

下载最新版按键精灵：https://www.123pan.com/s/zJGRVv-1FU0A，然后打开kernelSU管理器授权按键精灵超级用户权限，最后打开按键精灵会提示没有root权限。

### Expected behavior

由于无法使用脚本，只能降级到旧版本kernelsu，而无法使用新特性，对此我感到很遗憾。

### Logs

_No response_

### Device info

- Device:OnePlus Ace2 Pro
- OS Version:13
- KernelSU Version:0.9.4~1.0.1
- Kernel Version:5.15.74~5.15.123

### Additional context

_No response_

hamjin:
In v0.9.3 and v0.9.4, we replaced `vfs_statx` and `do_execveat_common` with syscall hooks. But we missed `fstatat64` and `compat_execve` and break compatibility with old and 32bit programs.

In one of the origin pull request(https://github.com/tiann/KernelSU/pull/1657), @tiann says "The compat_ series is not needed; because we don't support 32-bit ourselves; for the case of "kernel is 64-bit, but system is 32-bit", this situation did exist, but we have never provided a 32-bit manager, so it has never been supported, too." In fact, some 32bit binaries are still using compat syscalls to exec other programs.
image
image

In `kali`(NetHunter Terminal)  `__arm64_compat_sys_execve` is directly called, while `__arm64_sys_fstatat64` is called before it in JuiceSSH `libcom_sonelli_bash.so`(32bit bash-4.2) like the strace in #1813. 

So add these two hooks back to fix them.

Fixes #2045, #1813

### DIFF
--- a/website/docs/repos.json
+++ b/website/docs/repos.json
@@ -719,5 +719,12 @@
         "kernel_name": "android_kernel_samsung_sm7225",
         "kernel_link": "https://github.com/itejo443/android_kernel_samsung_sm7225",
         "devices": "Samsung Galaxy F23 and M23 (m23xq)"
+    },
+    {
+        "maintainer": "JuiIm",
+        "maintainer_link": "https://github.com/JuiIm",
+        "kernel_name": "M30s-custom-kernel",
+        "kernel_link": "https://github.com/JuiIm/M30s-custom-kernel-M307f---KernelSU-support/",
+        "devices": "Samsung Galaxy M30s (M307F)"
     }
 ]


tiann:
In order to be compatible with these applications that do not upgrade to 64-bit, we have to intercept these system calls on the devices of most users who probably won't even use these applications, which is very frustrating.

PR #2085, opened 2024-09-23
Update Magisk Version

CanerKaraca23:
Some modules demand Magisk 26+ from users, updated to latest.

Not tested.


tiann:
We don't need this anymore; the module needs to detect 'KSU' on its own. We upgraded to v26 once, and it caused problems.

PR #2096, opened 2024-09-26
api: spawn CallbackList use UiThreadHandler::runAndWait executor

whalechoi:
jsexception

Because the exit callback in js cleans up the ChildProcess, when the last callback of stdout is executed after the exit callback, an error like “emitData ReferenceError: spawn_callback_1727358276092_68 is not defined” will occur.


whalechoi:
image

This happens when spawn is used frequently, especially print a long line.

tiann (approved):

PR #2099, opened 2024-09-27
Fixed the app Lag and Non responcive issue on kernel su not installed devices

ALEX5402:
## Here is a demo video of before and after

# Before
https://github.com/user-attachments/assets/7bf3d3b7-d0c4-4b10-956a-de5499d2fb08

# After
https://github.com/user-attachments/assets/bae16e31-ec67-45ad-bcf8-eec196a47817

## Expected Behavior
- The app needs to be responsive.
- The app shouldn't be crashing.


tiann:
You can hide the menu, the menu is useless for unsupported device.

ALEX5402:
> You can hide the menu, the menu is useless for unsupported device.

got it good idea

PR #2102, opened 2024-09-30
Check for remote url instead of .git existence

hmtheboy154:
Current implementation of .git checking have some issues:
- If a kernel source being cloned using git-repo tool, especially newer version, it will clone .git in .repo inside of the submodule source.
- Newer kernel source changed how $(src) & $(srctree) output. In the case of kernel 6.11 as I tested $(src) show the relative path of the symlink in drivers/ instead of KernelSU main directory. Which makes checking for ../.git existence comletely useless.

One idea I would like to propose is to check for the remote URL in $(src). If the remote is correct then we can assume this is KSU repo and then count the version (also unshallow the repo). Here's what I did:
- Check if kernel is 6.9+ and then change $(src) accordingly.
- Print $(src) remote in verbose and check for the correct URL (both ssh & https).
- If correct, check if the repo is shallow by using git rev-parse --is-shallow-repository. Unshallow if it's return true.
- Calculate version as usual.

The only drawback I can see so far is that forks will have to edit the url in the Makefile, which I don't think it's much of an issue.


CanerKaraca23:
@hmtheboy154 There have been conflicts for a long time

dabao1955:
see https://github.com/tiann/KernelSU/pull/3108?

hmtheboy154:
> see #3108?

yea looks better, I'll close this one

PR #2156, opened 2024-10-23
manager: Add action.sh for user to manually trigger modules' functionality from manager

lingqiqi5211:
Commits:
- manager: Add `action.sh` for user to manually trigger modules' functionality from manager
- manager: Optimize ModuleItem
- manager: uninstall button: TextButton -> FilledTonalButton
- Optimize `run_action` function

Commit Author & Thank:
- @lightsummer233
- @lingqiqi5211
- [APatch](https://github.com/bmax121/APatch)

Demo Show:
image

### DIFF
--- a/userspace/ksud/src/module.rs
+++ b/userspace/ksud/src/module.rs
@@ -570,6 +570,28 @@ pub fn uninstall_module(id: &str) -> Result<()> {
     })
 }
 
+pub fn run_action(id: &str) -> Result<()> {
+    let action_script_path = format!("/data/adb/modules/{}/action.sh", id);
+    let result = Command::new(assets::BUSYBOX_PATH)
+        .args(["sh", &action_script_path])
+        .env("ASH_STANDALONE", "1")
+        .env(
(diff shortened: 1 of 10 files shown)


lightsummer233 (approved):

userspace/ksud/src/module.rs
```diff
@@ -570,6 +570,28 @@ pub fn uninstall_module(id: &str) -> Result<()> {
     })
 }
 
+pub fn run_action(id: &str) -> Result<()> {
```
Dr-TSNG, line 573: Why not use existing `exec_script`?

Dr-TSNG (changes requested):

userspace/ksud/src/module.rs
lingqiqi5211: I'm very sorry. I didn't notice that this function method exists. It has been optimized according to the situation.（qaq）

Dr-TSNG (approved):

PR #2189, opened 2024-11-04
new SearchAppBar ui

Houvven:
Preview

image

image


lightsummer233:
I don't think it good. 
It looks different from the other components

Houvven:
> I don't think it good. It looks different from the other components in material3

or this？
image

lightsummer233:
> > I don't think it good. It looks different from the other components in material3
> 
> or this？ image

Does it look fine with TopAppBar's scroll behavior?

PR #2203, opened 2024-11-11
Add oppo/oneplus/realme specific partitions

miuirussia:


tiann:
I don't think this is a good idea, for other partitions, mount it in `post-mount.sh` manually.

miuirussia:
How to mount partitions with overlayfs?

miuirussia:
@tiann

PR #2284, opened 2024-12-11
Magic Mount 

rifsxd:


LeeBinder:
Hi @rifsxd Can you shine a light on why you deleted the magic branch last month, and what the future of magic mount (vs. the modules.img sparse image) is for tiann/KernelSU from what you know?

PR #2305, opened 2024-12-21
Minor UI improvements

silvzr:
**Changes**
1. With the addition of the Action button the module card feels pretty bloated when all of them are present at the same time.
Screenshot_20241220-221644_KernelSU
To improve it we can use icons only instead of text since they take less space; this also means adding an icon for Update and Uninstall buttons.
Screenshot_20241220-220450_KernelSU
Below are displayed other possible behaviors when not all of them are present:
Screenshot_20241220-221227_KernelSU
Screenshot_20241220-221214_KernelSU
Screenshot_20241220-220428_KernelSU
Since the Update button appears less frequently than the others, make it always display its text unless both Action and WebUI are present.
Screenshot_20241220-221401_KernelSU

2. Minor change to avoid possible missclicks caused by smaller buttons. Enabling and disabling modules is pretty rare by itself and considering overlayfs isn't as fast as magic mount a missclick often feels pretty punishing.

3. The commit 7b3e732 readded transitions to all sections but I noticed they were missing from Home. With the help of @rifsxd I found out it was cause of the lag generated by the install function, which is called everytime you go to Home. Since installing magiskboot multiple times isn't needed, moving it to MainActivity ensures it's executed on app launch only avoiding unnecessary lag.

Tested-by: backslashxx 118538522+backslashxx@users.noreply.github.com

### DIFF
--- a/manager/app/src/main/java/me/weishu/kernelsu/ui/MainActivity.kt
+++ b/manager/app/src/main/java/me/weishu/kernelsu/ui/MainActivity.kt
@@ -45,6 +45,7 @@ import me.weishu.kernelsu.ui.screen.BottomBarDestination
 import me.weishu.kernelsu.ui.theme.KernelSUTheme
 import me.weishu.kernelsu.ui.util.LocalSnackbarHost
 import me.weishu.kernelsu.ui.util.rootAvailable
+import me.weishu.kernelsu.ui.util.install
 
 class MainActivity : ComponentActivity() {
 
@@ -57,6 +58,9 @@ class MainActivity : ComponentActivity() {
         }
 
         super.onCreate(savedInstanceState)
+
+	val isManager = Natives.becomeManager(ksuApp.packageName)
+	if (isManager) install()
 
         setContent {
             KernelSUTheme {
--- a/manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
+++ b/manager/app/src/main/java/me/weishu/kernelsu/ui/screen/Home.kt
@@ -73,9 +73,6 @@ fun HomeScreen(navigator: DestinationsNavigator) {
             verticalArrangement = Arrangement.spacedBy(16.dp)
         ) {
             val isManager = Natives.becomeManager(ksuApp.packageName)
-            SideEffect {
-                if (isManager) install()
-            }
             val ksuVersion = if (isManager) Natives.version else null
             val lkmMode = ksuVersion?.let {
                 if (it >= Natives.MINIMAL_SUPPORTED_KERNEL_LKM && kernelVersion.isGKI()) Natives.isLkmMode else null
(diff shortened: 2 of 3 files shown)


backslashxx:
heres how it looks for me
image
image

Tested-by: backslashxx 118538522+backslashxx@users.noreply.github.com

lightsummer233:
I don’t think only shows the icon will looks nice 
It looks weird, we needn’t too much icons if we have spaces to display texts
And the buttons which only contains a icon is too small 
u can adjust the spacer between buttons i think

silvzr:
> I don’t think only shows the icon will looks nice 
> In many times, we will only see less than 3 buttons 
> We needn’t too much icons if we have spaces to display texts
> And the buttons which only contains a icon is too small 
> u can adjust the spacer between buttons i think 

Well I can edit it to display text is there's only another one button but question is, how many times do you actually need to do it? Missclicking isn't punishing anymore partly for this reason. Also displaying both text and icon when update and uninstall are present would feels disproportionate on the left side. Spacing cannot be improved much, depending on DPI you could have even less space with the actual spacing and I have examples for that

lightsummer233:
> > I don’t think only shows the icon will looks nice 
> > In many times, we will only see less than 3 buttons 
> > We needn’t too much icons if we have spaces to display texts
> > And the buttons which only contains a icon is too small 
> > u can adjust the spacer between buttons i think 
> 
> Well I can edit it to display text if there's only another one button but question is, how many times do you actually need to do it? Missclicking isn't punishing anymore partly for this reason. Also displaying both text and icon when update and uninstall are present would feels disproportionate on the left side (maybe make it display text only with action or webui, but not with update?). Spacing cannot be improved much, depending on DPI you could have even less free space with the actual one - here's an example
> 
> 

If u want both ends to be proportional and have more spaces, u are right
Display the icon only is a good way
Btw I think we'd better to add some content description that can be shown by long pressing the button, it can help users to understand

silvzr:
> > > I don’t think only shows the icon will looks nice
> > > In many times, we will only see less than 3 buttons
> > > We needn’t too much icons if we have spaces to display texts
> > > And the buttons which only contains a icon is too small
> > > u can adjust the spacer between buttons i think
> > 
> > 
> > Well I can edit it to display text if there's only another one button but question is, how many times do you actually need to do it? Missclicking isn't punishing anymore partly for this reason. Also displaying both text and icon when update and uninstall are present would feels disproportionate on the left side (maybe make it display text only with action or webui, but not with update?). Spacing cannot be improved much, depending on DPI you could have even less free space with the actual one - here's an example
> > 
> 
> If u want both ends to be proportional and have more spaces, u are right Display the icon only is a good way Btw I think we'd better to add some content description that can be shown by long pressing the button, it can help users to understand

I tried adding tooltips but
- making them display when text is present would be useless, and defining conditions for them to appear would mean that other 'futile' ifs are added just for that. Icon are pretty recognizable imo, as I said the least one is update which always display text unless there are all 4 at the same time.
- importing it from compose seems to trigger unresolved reference for positionProvider even if TooltipDefaults is imported too. I may be wrong but it's available only since M3 1.4.0-alpha-01 
If you can manage to code it avoiding these two issues feel free to do so

tiann (approved):

PR #2306, opened 2024-12-22
fix docs:there is no ksud in releases so change it into actions

model-err:
I have changed all the documentations that I know have ksud usages&download links,other languages I think has no ksud usages.
我把我能找到的ksud下载链接都改了,其他的translation我怀疑没有写ksud的使用及下载...


tiann:
We should fix the release, not the docs.

PR #2307, opened 2024-12-22
Imitate terminal clear command in manager

bryanyee33:
The current implementation of KSU manager's output screen simply prints `[H[J` when the `clear` command is used (in both the flashing module & action button screen) instead of clearing the screen:

This limits the ability of shell scripts to purely textual & linear outputs, and prevents more flexible outputs such as a refreshing progress bar or even a progress circle for long running scripts. The current implementation moreover limits the output to 65536 bytes for the String `text`, causing the app to hang once this limit has been reached for scripts with more verbose outputs.

This PR fixes these issues by allowing for usage of the `clear` command in shell scripts to clear the screen. It works by checking if the current output line starts with "[H[J", which is the default output of the `clear` command in KSU's busybox, and clears the previous outputs if there is a match. This should work universally since the `clear` command defaults to this implementation when ran in KSU manager.

A working example can be seen below, where the `clear` command is heavily used (24 times a second) to test for performance & reliability of the code:

https://github.com/user-attachments/assets/c45fb6f1-1b40-4b67-8837-4d9a00429715

Tested-by: backslashxx

### DIFF
--- a/manager/app/src/main/java/me/weishu/kernelsu/ui/screen/ExecuteModuleAction.kt
+++ b/manager/app/src/main/java/me/weishu/kernelsu/ui/screen/ExecuteModuleAction.kt
@@ -49,6 +49,7 @@ import java.util.Locale
 @Destination<RootGraph>
 fun ExecuteModuleActionScreen(navigator: DestinationsNavigator, moduleId: String) {
     var text by rememberSaveable { mutableStateOf("") }
+    var tempText : String
     val logContent = rememberSaveable { StringBuilder() }
     val snackBarHost = LocalSnackbarHost.current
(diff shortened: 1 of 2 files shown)


backslashxx:
https://github.com/user-attachments/assets/05e38949-222c-42b6-90f7-0a1005dbaa8d

here on my side, works well !

tiann (approved):

backslashxx:
so, 11998 tentatively?
```
[ "$KSU_VER_CODE" -ge 11998 ] && clear
```

backslashxx:
here was the demo just for logging purposes
https://github.com/backslashxx/bad-apple-ascii/tree/magisk-module

PR #2381, opened 2025-01-25
Avoid unnecessarily overriding capabilities

aviraxp:
Previously all capabilities were overridden with 0xffffffffffffffff, which is
not what normal processes have. This causes PR_CAPBSET_DROP prctl
fail because it is dropping invalid caps. Fix it.

This fixes https://gitlab.com/kalilinux/nethunter/apps/kali-nethunter-app/-/issues/378.

Co-Authored-By: 5ec1cff <56485584+5ec1cff@users.noreply.github.com>

### DIFF
--- a/kernel/allowlist.c
+++ b/kernel/allowlist.c
@@ -1,3 +1,4 @@
+#include <linux/capability.h>
 #include <linux/compiler.h>
 #include <linux/fs.h>
 #include <linux/gfp.h>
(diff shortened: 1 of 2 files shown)


aviraxp:
Wrong, need to avoid hardcoding caps value, maybe via get_task_cred to get init's cap

tiann (approved):

tiann (approved):

PR #2386, opened 2025-01-26
Clean up umount targets

aviraxp:


aviraxp:
I am still not sure why MNT_DETACH cannot be used for overlayfs mountpoints

backslashxx:
any chance that this list be extended on ksud?

like for example 
```
ksud try_umount pm path com.google.android.youtube | sed 's/package://'
```

tiann (approved):

tiann:
> I am still not sure why MNT_DETACH cannot be used for overlayfs mountpoints

On my test device, using MNT_DETACH, umount may return -39.

backslashxx:
> On my test device, using MNT_DETACH, umount may return -39.

@tiann I have this working, but I don't understand why it has to be as complex as this

https://github.com/tiann/KernelSU/compare/cfbbded%E2%80%A67e34953

PR #2458, opened 2025-02-20
Provide TMPDIR for boot image repacking

aviraxp:
/data/local/tmp is never writable for normal apps, why previously it works is that Rust's temp_dir() gets path from env, and since A13, TMPDIR is set to app's cache dir. This is not the case for A12, so it breaks. Fix it by set TMPDIR ourselves.


aviraxp:
need test

aviraxp:
works

tiann (approved):

PR #2475, opened 2025-02-25
[RFC] kernel: restore non-kprobe support, sucompat toggle for non-kp.

backslashxx:
this will revert 500ff9b
"kernel: remove unused CONFIG guard becuase GKI kernel enable kprobe by default"

This pr can allow gki kernels to be built with manual hooks. It also comes with a port of 
sucompat toggle for the sake of feature parity.

If this gets accepted, I also plan to get manual hooks guide on par with kprobes hook.
https://github.com/backslashxx/KernelSU/issues/5

ofcourse, this will depend now on KernelSU policy, as this will **re-introduce an ifdef hell**. 
This will make maintenance harder, you now have to keep track a lot of different states, etc etc,
but even if this does NOT get accepted, this can still be a demo on how you can have manual 
hooks on gki kernels.

### DIFF
--- a/kernel/Kconfig
+++ b/kernel/Kconfig
@@ -2,6 +2,7 @@ menu "KernelSU"
 
 config KSU
 	tristate "KernelSU function support"
+	depends on EXT4_FS
 	depends on OVERLAY_FS
 	default y
 	help
--- a/kernel/Kconfig
+++ b/kernel/Kconfig
@@ -9,6 +9,14 @@ config KSU
 	  To compile as a module, choose M here: the
 	  module will be called kernelsu.
 
+config KSU_WITH_KPROBES
+	bool "Use kprobes for kernelsu"
+	depends on KSU
+	depends on KPROBES
+	default y
+	help
+	  Make sure to disable this if you use manual hooks.
+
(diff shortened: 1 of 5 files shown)


ChiseWaguri:
+1, manual hooks are useful for both GKI and non-GKI users. Non-GKI users benefit from compatibility, while GKI users get an alternative. Supporting both keeps the flexibility, even if it turns into an #ifdef maze, just having the option is huge. I've been using manual hooks, even tho I am a gki user. ChiseWaguri/android_kernel_xiaomi_marble@3322eec

hazepynut:
> +1, manual hooks are useful for both GKI and non-GKI users. Non-GKI users benefit from compatibility, while GKI users get an alternative. Supporting both keeps the flexibility, even if it turns into an #ifdef maze, just having the option is huge. I've been using manual hooks, even tho I am a gki user. [ChiseWaguri/android_kernel_xiaomi_marble@3322eec](https://github.com/ChiseWaguri/android_kernel_xiaomi_marble/commit/3322eec)

+1

tiann:
Just reverting is fine; we don't want any extra complexity.

PR #2513, opened 2025-03-15
ksud/installer: /odm handling

backslashxx:
we move the folder out of system if it exists in real filesystem and it is not a symlink.
this is already supported on init_event.rs so only handle_partition logic was needed to make it happen

since KernelSU is using overlayfs, we need to move these out.


backslashxx:
@tiann :point_right: :point_left:

PR #2517, opened 2025-03-18
Add a .gitignore file to the kernel directory

fnr1r:
This PR is to fix a personal annoyance. The `.o` and `.o.cmd` files generated during the build process should probably be ignored by Git.

Also @KernelSU-Next or @rifsxd, please pull this as well. (if this gets accepted)

Thank you.

### DIFF
--- a/kernel/.gitignore
+++ b/kernel/.gitignore
@@ -1,0 +1,3 @@
+modules.*
+*.o
+*.o.cmd


rifsxd:
It probably won't get merged because i dont think the objects doesn't even get compiled in the kernel dir

tiann:
You should configure your build system and specify the objects directory.

PR #2531, opened 2025-03-27
kernel: core_hook: automate and refactor umount

backslashxx:
monitors and lists all incoming mounts by hooking **security_sb_mount** or alternatively, **sys_mount / do_mount**. 
this requires ksud to go back and use sys_mount instead of fsopen + move_mount

this also fixes that MNT_DETACH issue. (https://github.com/tiann/KernelSU/pull/2386#issuecomment-2615518282)
Modules also can create an overlay with KSU devname and it will be added to the list.
**This will also restore pre 5.2 manager support.**

Those might be a bit questionable, so it is a take it or reject it kind of situation.

sys_mount hook looks like
```
ksu_mount_monitor(copy_mount_string(dev_name), copy_mount_string(dir_name), copy_mount_string(type));
```
ref: https://github.com/backslashxx/mojito_krenol/commit/ac020b09d0ff7a02dea9d1411202d30b5806ec14

sys_mount / do_mount compatibility is for LKM and 6.8+ where you cant just do LSM hooks as is.
https://github.com/torvalds/linux/commit/9285c5ad9d00abfe0f4e2ce4039c8127e7a09738#diff-3b35df76d1c4627ed97cc307a957850d57b85e77917c19ac0813286c46b4ace2R556

Context: https://github.com/aviraxp/KernelSU/commit/c7facef72a6ac6bdd6039b20b7df9ec3ea511331#commitcomment-153941502

### DIFF
--- a/kernel/core_hook.c
+++ b/kernel/core_hook.c
@@ -704,10 +704,13 @@ static int ksu_task_fix_setuid(struct cred *new, const struct cred *old,
 static int ksu_sb_mount(const char *dev_name, const struct path *path,
                         const char *type, unsigned long flags, void *data)
 {
-	char buf[256];
+	// 384 is what throne_tracker uses, something sensible even for /data/app
+	// we can pattern match revanced mounts even.
+	// we are not really interested on mountpoints that are longer than that
+	char buf[384];
 	char *dir_name = d_path(path, buf, sizeof(buf));
-	
-	if (dir_name)
+
+	if (dir_name && dir_name != buf)
 		return ksu_mount_monitor(dev_name, dir_name, type);
 	else
 		return 0;
--- a/kernel/core_hook.c
+++ b/kernel/core_hook.c
@@ -559,6 +559,7 @@ static void try_umount(const char *mnt, bool check_mnt, int flags)
 
 int ksu_handle_setuid(struct cred *new, const struct cred *old)
 {
+	char buf[256];
 	struct mount_entry *entry, *tmp;
 
 	// this hook is used for umounting overlayfs for some uid, if there isn't any module mounted, just ignore it!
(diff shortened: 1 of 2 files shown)


backslashxx:
yes it is incomplete, it doesnt have the kprobe hook, but I cant really test this.

tiann:
What is the root cause of https://github.com/tiann/KernelSU/pull/2386#issuecomment-2615518282, and how does it fix the issue?

backslashxx:
> What is the root cause of [#2386 (comment)](https://github.com/tiann/KernelSU/pull/2386#issuecomment-2615518282), and how does it fix the issue?

That I don't really have an idea, I suspect it has something to do with mount order and then unmount order.

We have to follow
mount: A B C D
unmount: D C B A

TheSakshamSingla (approved):

kernel/core_hook.c
```diff
@@ -332,6 +333,9 @@ int ksu_handle_prctl(int option, unsigned long arg2, unsigned long arg3,
 			if (!boot_complete_lock) {
 				boot_complete_lock = true;
 				pr_info("boot_complete triggered\n");
+				// turn off mount monitor
```
aviraxp, line 336: 1）post-fs-data can get triggered again after boot complete (userspace reboot)
2)  I am not sure, if i clone init's ns after boot, will the new mounts get monitored?

kernel/core_hook.c
```diff
@@ -707,7 +707,13 @@ static int ksu_task_fix_setuid(struct cred *new, const struct cred *old,
 static int ksu_sb_mount(const char *dev_name, const struct path *path,
                         const char *type, unsigned long flags, void *data)
 {
-	return ksu_mount_monitor(dev_name, path, type);
+	char buf[256];
```
aviraxp, line 718: possible buffer overflow?

kernel/core_hook.c
backslashxx: no idea with this, I can retest later, though, this hook turn off is purely optional, you can keep it open.

sys_mount needs priv so its not like you can do detections with this.

kernel/core_hook.c
backslashxx: we can go PATH_MAX if you want

backslashxx: we can use d_path's return value if it fails
```
* Returns a pointer into the buffer or an error code if the path was too long
```

ah yeah fixed this, whats a good size for it though?

for normal ksu mounts, they arent even longer than 64 chars, but if you want to reuse it for something then yeah

backslashxx:
I'm not really sure on the kprobe hook

PR #2584, opened 2025-05-15
kernel/throne_tracker: we just uninstalled the manager, stop looking for it

backslashxx:
When the manager UID disappears from packages.list, we correctly
invalidate it — good. But, in the very next breath, we start scanning
/data/app hoping to find it again?

This event is just unnecessary I/O, exactly when we should be doing less.
Apparently this causes hangups and stuckups which is REALLY noticeable
on Ultra-Legacy devices.

Skip the scan — we’ll catch the reinstall next time packages.list updates.

### DIFF
--- a/kernel/throne_tracker.c
+++ b/kernel/throne_tracker.c
@@ -357,12 +357,14 @@ void track_throne()
 		if (ksu_is_manager_uid_valid()) {
 			pr_info("manager is uninstalled, invalidate it!\n");
 			ksu_invalidate_manager_uid();
+			goto prune;
 		}
 		pr_info("Searching manager...\n");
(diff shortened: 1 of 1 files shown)


tiann (approved):

backslashxx:
Thanks. We even had to put this on a kthread man. (Linux 3.4 / 3.10)

https://github.com/tiann/KernelSU/commit/9658e23
https://github.com/tiann/KernelSU/commit/5929026
https://github.com/ximi-libra-test/android_kernel_xiaomi_libra/commits/lineage-18.1-ksu/
https://github.com/acroreiser/android_kernel_lge_hammerhead/commits/lineage-22.2-kernelsu/

PR #2591, opened 2025-05-17
ci: update kmi versions

ukriu:


ukriu:
Also check Kernel-SU/AnyKernel3#5

tiann (approved):

PR #2595, opened 2025-05-18
kernel: ksud: dont create structs just to deconstruct them for a string

backslashxx:


backslashxx:
noticed this after trying to do integration on Linux 3.0

tiann:
> This also makes integration easier on kernels that don’t have struct filename.

Which kernels lack this struct?

backslashxx:
ah yeah, that is actually irrelevant for current ksu, but this is introduced on 3.7-rc1

tiann:
This is not good, it transforms an explicit dependency into an implicit one.

backslashxx:
which? this is the only part on kprobes that needs filename struct, execve_sucompat hook doesnt use it already, you might as well free this part from struct filename use

tiann:
> This is not good, it transforms an explicit dependency into an implicit one.

So, why should we care about pre-3.7 kernels?

backslashxx:
you dont have to, but yeah this is still creating a dummy struct that it just deconstructs. 
since this is mostly only for UL, it is very okay to reject this

tiann:
The first member of the struct filename is `name`, so the pointer to the struct points to `name`. This creates an implicit dependency. Although it may remain the same indefinitely, any change will cause a panic. The benefits apply only to pre-3.7 kernels, making it not worth the effort.

backslashxx:
isnt the flow like

sys_execve --> kprobe --> sys_execve_handler_pre --> create dummy struct filename --> ksu_handle_execveat_ksud

now itll be like

sys_execve --> kprobe --> sys_execve_handler_pre --> ksu_handle_execveat_ksud

tiann:
the struct is allocated on stack, which is very cheap.

PR #2600, opened 2025-05-21
ksud: add /my_bigball partition support

chisewaguri:
some OnePlus devices mount a nonstandard partition at /my_bigball, which is used similarly to /odm or /vendor for preloaded system components.

this commit:
- adds /my_bigball to the partition list in init_event.rs
- handles /my_bigball in installer.sh like other partitions

### DIFF
--- a/kernel/core_hook.c
+++ b/kernel/core_hook.c
@@ -605,6 +605,7 @@ int ksu_handle_setuid(struct cred *new, const struct cred *old)
 
 	// fixme: use `collect_mounts` and `iterate_mount` to iterate all mountpoint and
 	// filter the mountpoint whose target is `/data/adb`
+	try_umount("/my_bigball", true, 0);
 	try_umount("/odm", true, 0);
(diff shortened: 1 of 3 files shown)


tiann:
Please use `post-mount.sh` in modules.

chisewaguri:
> Please use `post-mount.sh` in modules.
we wouldn't get to use the umount feature that way :'

I mean yeah, I get not wanting to extend the mount logic for every weird partition some oem dreams up, but OnePlus is actually using it like /odm or /vendor, preloading system components into it.

if we already handle /odm, /vendor, /product, etc., what's one more?

tiann:
There are many devices in the world, and each may have a special partition. Should we add it here? It would be a complete mess.

chisewaguri:
fair enough

backslashxx:
Theres a way to get this umount

https://github.com/tiann/KernelSU/pull/2531

with this, just set devname as KSU and itll be added to its umount list

PR #2626, opened 2025-06-08
throne_tracker: avoid cross fs access

aviraxp:
Files in /data/app may be stacked on incremental fs, if user installs big apps from play store or adb shell. Performing I/O operation on it may results in long-time blocking. As KSU won't get installed in those ways, just avoid cross fs access.


backslashxx:
missing path_put

PR #2631, opened 2025-06-10
Switch to prepare_creds/commit_creds

aviraxp:
Update API as per kernel doc recommends, also fix setup_groups refcount leak while at it.

### DIFF
--- a/kernel/core_hook.c
+++ b/kernel/core_hook.c
@@ -108,6 +108,7 @@ static void setup_groups(struct root_profile *profile, struct cred *cred)
 
 	groups_sort(group_info);
 	set_groups(cred, group_info);
+	put_group_info(group_info);
 }
 
 static void disable_seccomp()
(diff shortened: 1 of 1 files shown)


kernel/core_hook.c
```diff
@@ -132,18 +133,18 @@ void escape_to_root(void)
...
+	cred = prepare_creds();
+	if (!cred) {
+		pr_warn("prepare_creds failed!\n");
+		return;
+	}
 
 	if (cred->euid.val == 0) {
 		pr_warn("Already root, don't escape!\n");
-		rcu_read_unlock();
+		abort_creds(cred);
```
backslashxx, line 144: all good even with 3.10, boots, werks, I has su.

move this check at entry, say check `current_euid().val == 0`
just to micro-optimize and avoid that extra logic.
https://github.com/ximi-libra-test/android_kernel_xiaomi_libra/commit/cd1320c

image

selfmusing: Do we have to revert this backport from now :
https://github.com/selfmusing/kernel_xiaomi_violet/commit/de9601f2fd646c9e3eb5b0ecbdadab14ee7b12cd ? 👀

PR #2632, opened 2025-06-11
kernel: throne_tracker: offload to kthread

backslashxx:
```
Run throne_tracker() in kthread instead of blocking the caller.
Prevents full lockup during installation and removing the manager.

First run remains synchronous for compatibility purposes (FDE, FBEv1, FBEv2)

Features:
- looks and waits for manager UID in /data/system/packages.list
- run track_throne() in a kthread after the first synchronous run
- prevent duplicate thread creation with a single-instance check
- spinlock-on-d_lock based polling adressing possible race conditions.

Race conditions adressed
- single instance kthread lock, smp_mb()
- track_throne_function, packages.list, spinlock-on-d_lock based polling
- is_manager_apk, apk, spinlock-on-d_lock based polling

This is a squash of:
kernel: throne_tracker: move throne_tracker to kthread with spinlocks
kernel: throne_tracker: harden track_throne_function file read
kernel: throne_tracker, apk_sign: functionify d_lock spinlock check
kernel: throne_tracker: harden packages.list checker further
kernel: apk_sign: loop file open on is_manager_apk
kernel: throne_tracker: tweak search_manager

Original skeleton based on:
`kernelsu: move throne_tracker() to kthread`
`kernelsu: check locking before accessing files and dirs during searching manager`
`kernelsu: look for manager UID in /data/system/packages.list, not /data/system/packages.list.tmp`
https://github.com/acroreiser/android_kernel_lge_hammerhead/compare/0b05e927...8783badd

Co-Authored-By: backslashxx <118538522+backslashxx@users.noreply.github.com>
Co-Authored-By: Yaroslav Zviezda <10716792+acroreiser@users.noreply.github.com>
```

### DIFF
--- a/kernel/apk_sign.c
+++ b/kernel/apk_sign.c
@@ -187,6 +187,24 @@ static __always_inline bool check_v2_signature(char *path,
 	bool v3_1_signing_exist = false;
 
 	int i;
+	struct path kpath;
+	if (kern_path(path, 0, &kpath))
+		return false;
+
+	// probably wont happen, just to be sure
+	if (!kpath.dentry) {
+		path_put(&kpath);
+		return false;
+	}
+
+	if (!spin_trylock(&kpath.dentry->d_lock)) {
+		pr_info("%s: lock held, bail out!\n", __func__);
+		path_put(&kpath);
+		return false;
+	}
+	spin_unlock(&kpath.dentry->d_lock);
+	path_put(&kpath);
+
 	struct file *fp = ksu_filp_open_compat(path, O_RDONLY, 0);
 	if (IS_ERR(fp)) {
 		pr_err("open %s error.\n", path);
--- a/kernel/apk_sign.c
+++ b/kernel/apk_sign.c
@@ -188,10 +188,6 @@ static __always_inline bool check_v2_signature(char *path,
 	bool v3_1_signing_exist = false;
 
 	int i;
-
-	if (is_lock_held(path))
-		return false;
-
 	struct file *fp = ksu_filp_open_compat(path, O_RDONLY, 0);
 	if (IS_ERR(fp)) {
 		pr_err("open %s error.\n", path);
@@ -321,5 +317,21 @@ module_param_cb(ksu_debug_manager_uid, &expected_size_ops,
 
 bool is_manager_apk(char *path)
 {
+	int tries = 0;
+
+	while (tries++ < 10) {
+		if (!is_lock_held(path))
+			break;
+
+		pr_info("%s: waiting for %s\n", __func__, path);
+		msleep(100);
+	}
+
+	// let it go, if retry fails, check_v2_signature will fail to open it anyway
+	if (tries == 10) {
+		pr_info("%s: timeout for %s\n", __func__, path);
+		return false;
+	}
+
 	return check_v2_signature(path, EXPECTED_SIZE, EXPECTED_HASH);
 }
--- a/kernel/apk_sign.c
+++ b/kernel/apk_sign.c
@@ -17,6 +17,7 @@
 #include "apk_sign.h"
 #include "klog.h" // IWYU pragma: keep
 #include "kernel_compat.h"
+#include "throne_tracker.h"
 
 
 struct sdesc {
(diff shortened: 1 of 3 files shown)


backslashxx:
context:

we are working on proper downstream support of MKSU on 3.X kernels.
we had a lot of stuckups/crashes when throne tracker runs, (manager uninstall, reinstall, install)

kthreaded throne tracker is based on the work and advice of @acroreiser (@LineageOS-UL)
initially, we tried to use a [full fat sleepable one](https://github.com/backslashxx/KernelSU/commit/9658e23594f26f74c6378d64a3e57194df02167f), but theres that issue where you can  `ps | grep throne` 
and it is too "fat" so we skipped it.

The one in here is a use and destroy one. 
_ofcourse for detection specialists, yes theres a pid increase everytime throne_tracker runs :^)_

I believe that this is useful to send upstream (here). While the code kind of targeted 3.X kernels,
with the proper choices on here that is as generic as possible, this should work as is even on 6.X.

selfmusing:
I've just tried this PR .
At first everything looks good and works fine as well , but after manually uninstalling and reinstalling the manager a few times it shows this error : 
Screenshot_20250611-130118_KernelSU

Crowning takes place as it should  but it has no effect on manager :

```
[  621.832459] KernelSU: renameat: packages.list.tmp -> packages.list, new path: /system/packages.list
[  621.833171] KernelSU: throne_tracker_thread: pid: 22618 started
[  621.833450] KernelSU: track_throne_function: /data/system/packages.list found!
[  621.913086] KernelSU: Searching manager...
[  621.916076] KernelSU: sha256: d06917e6deece7bf9b3278b1bb4ca4123c7d546bfebfd204915465804750bd03, expected: d06917e6deece7bf9b3278b1bb4ca4123c7d546bfebfd204915465804750bd03
[  621.916175] KernelSU: Found new base.apk at path: /data/app/~~v5qVNzdlgF8NhZAORbb8Ww==/me.weishu.kernelsu-lLxyXj-GG2N6QSVPSXMPGg==/base.apk, is_manager: 1
[  621.916178] KernelSU: manager pkg: me.weishu.kernelsu
[  621.916254] KernelSU: Crowning manager: me.weishu.kernelsu(uid=10579)
[  621.916279] KernelSU: Stop searching
[  621.916390] KernelSU: Search manager finished
[  621.917127] KernelSU: throne_tracker_thread: pid: 22618 exit!
```

I had to reinstall it again several times to fix the signature hash mismatch on ksu manager.

backslashxx:
yeah there can be an issue with that hash managemnent
I had a workaround for this, lemem push.

selfmusing:
> yeah theres can be an issue with that hash managemnent I had a workaround for this, lemem push.

Thanks! It fixed the issue.

backslashxx:
yeah we'll try another approach once that build of yours compiles @selfmusing

oct888:
I've just tried this PR, Everything seems good on my end using 5.15. manager successfully crowned across multiple uninstall/reinstall cycles.

tiann:
I understand that this PR has two features:

1. Switch to kthread
2. Use s_magic to replace super_block

Please split it into two PRs.

backslashxx:
> I understand that this PR has two features:
> 
> 1. Switch to kthread
> 2. Use s_magic to replace super_block
> 
> Please split it into two PRs.

avira said so that I just push both, but sure, I will push cross-fs-avoidance first.
wait a little.

aviraxp:
My idea is to vmalloc() package.list.tmp to memory, use vfs_read() to read it, create kthread to parse it and search. Or we are guaranteed to have a race.

backslashxx:
now thats something above my paygrade, I can lookup about it later.
but yeah my current dumb check can cause a 1 second stuck time, I say its pretty cheap as were now fully async.

--
as for the lock, the spinlock there being unlocked right away is just a check for d_lock to prevent processing a file being renamed / deleted etc, whatever operation holds a d_lock

this likely is optional once packages.list is fully committed, so I'm not really sure about this.

we can hold this lock longer on apk_sign to assure proper processing.

we can also try to use this as logic to check for packages.list if its unlocked.

--
as for this being designed for 3.X , this already has been built and tested on @ShirkNeko 's  SUKiSU, both with GKI and LKM
so it should build fine 5.10++

https://github.com/SukiSU-Ultra/SukiSU-Ultra/actions

backslashxx:
@aviraxp theres also another race on having multiple kthreads.
I literally saw it happen on an SD 808 of all things.

right now I am putting smp_mb() on it, but this might be too much.

current behavior @ 8717a4da0d0f97680a2fa685e1efa3c766bf726e
https://paste.debian.net/plain/1379323

backslashxx:
ig with the looping spinlock check, is_lock_held might be better renamed to is_file_unstable

tiann:
Besides supporting the 3.X kernel, are there any other benefits to this patch? I don't understand why it needs to loop 10 times; my intuition tells me something feels off about this.

backslashxx:
well its not really about supporting 3.x, its about not making that rename blocking
avira is doing something with those two races on packages.list.tmp -> packages.list rename

he said it can be an issue on LKM users as they see the rename earlier

--

it actually already works now, its all about hardening around those races

aviraxp:
> Besides supporting the 3.X kernel, are there any other benefits to this patch? I don't understand why it needs to loop 10 times; my intuition tells me something feels off about this.

Note that searching manager now is async. So we are not blocking renameat anymore. So there will be a race when searching manager, renameat hasn't finished at all. I think we can move renameat hook to post instead of pre.

But there will be another race if two packages are installed in a short time, like 10ms.

backslashxx:
so with toying around theres three weak points

1. that single instance kthread lock
2. opening of packages.list
3. opening of apk itself on check_v2_signature

-
1. is easy to solve but a bit overkill
2. I have a dumb loop
3. same as 2.

PR #2633, opened 2025-06-12
kernel: throne_tracker: avoid cross-fs traversal using s_magic check

backslashxx:
Skip directories that does NOT have the same magic as /data/app.
This is to avoid scanning incfs and any other stacked filesystems.

While this is way dumber, it's way cheaper.
no kern_path(), no missable path_put(), no ref handling.

This supercedes
`throne_tracker: avoid cross fs access (https://github.com/tiann/KernelSU/pull/2626)`
- upstream https://github.com/tiann/KernelSU/commit/0b6998b474ed00610bbf2d6679f853cef07af3b8

### DIFF
--- a/kernel/throne_tracker.c
+++ b/kernel/throne_tracker.c
@@ -213,7 +213,8 @@ void search_manager(const char *path, int depth, struct list_head *uid_data)
 	int i, stop = 0;
 	struct list_head data_path_list;
 	INIT_LIST_HEAD(&data_path_list);
-
+	static unsigned long data_app_magic __read_mostly = 0;
+	
 	// Initialize APK cache list
 	struct apk_path_hash *pos, *n;
 	list_for_each_entry(pos, &apk_path_hash_list, list) {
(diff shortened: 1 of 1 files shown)


backslashxx:
@tiann 
https://github.com/tiann/KernelSU/pull/2632#issuecomment-2965047621

heres that "simpler cross-fs avoidance"

once this is in and then tested, I can re-prep kthread

kernel/throne_tracker.c
```diff
@@ -272,6 +245,22 @@ void search_manager(const char *path, int depth, struct list_head *uid_data)
...
 					goto skip_iterate;
 				}
+				
+				// grab magic on first folder, which is /data/app
+				if (unlikely(!data_app_magic)) {
+					if (file->f_inode->i_sb->s_magic) {
+						data_app_magic = file->f_inode->i_sb->s_magic;
+						pr_info("%s: dir: %s got magic! 0x%lx\n", __func__, pos->dirpath, data_app_magic);
+					} else
+						goto skip_iterate;
```
tiann, line 256: filp_close(file)

tiann (changes requested):

kernel/throne_tracker.c
backslashxx: yup, nice catch

tiann (dismissed):

backslashxx:
for reference purposes, this change adds this

image

kernel/throne_tracker.c
```diff
@@ -226,19 +210,11 @@ FILLDIR_RETURN_TYPE my_actor(struct dir_context *ctx, const char *name,
...
 	INIT_LIST_HEAD(&data_path_list);
-
-	err = kern_path(path, 0, &kpath);
-
-	if (err) {
-		pr_err("get search root %s err: %d\n", path, err);
-		return;
-	}
-
+	static unsigned long data_app_magic __read_mostly = 0;
```
aviraxp: Consider:
1) Phone boot up, data_app_magic is ext4 magic (for example);
2) Search manager complete;
3) User mount overlayfs on /data/app;
4) Search manager again;
5) As data_app_magic is static, it will keep ext4 value;
6) All following operation will fail.

backslashxx: ah yeah that was intended, I didnt know theres that overlayfs over /data/app usecase.
I guess with casefolded ofs support patches thats doable. wait a little.
I also pray that that thing gets merged upstream.

aviraxp (approved):

PR #2653, opened 2025-06-25
kernel: ksud, core_hook: migrate ksud execution to  security_bprm_check

backslashxx:
```
This migrates ksud execution decision-making to bprm_check_security.
This requires passing proper argv and envp to a modified _ksud handler
aptly named 'ksu_handle_bprm_ksud'.

Introduces:
int ksu_handle_bprm_ksud(const char *filename, const char *argv1, 
			const char *envp, size_t envp_len)


which is adapted from:
int ksu_handle_execveat_ksud(int *fd, struct filename **filename_ptr,
			     struct user_arg_ptr *argv,
			     struct user_arg_ptr *envp,
			     int *flags)

ksu_handle_bprm_ksud handles all the decision making, it decides when it is
time to apply_kernelsu_rules depending if it sees "second_stage".

For LSM hook, turns out we can pull out argv and envp from mm_struct.
The code in here explains itself on how to do it.

whole blob exists on arg_start to arg_end, so we just pull it out and grab next
array after the first null terminator.

as for envp, we pass the pointer then hunt for it when needed

My reasoning on adding a fallback on usercopy is that on some devices a fault
happens, and it copies garbled data. On my creation of this, I actually had to lock
that  _nofault copy on a spinlock as a way to mimic preempt_disable/enable
without actually doing it. As per user reports, no failed _nofault copies anyway
but we have-to-have a fallback for resilience.

References:
- old version https://github.com/backslashxx/KernelSU/commit/6efcd8193e62d13a4e62cda0ce54d6770260c680
- bad usercopy https://github.com/backslashxx/KernelSU/issues/21

This now provides a small helper function, ksu_copy_from_user_retry, which explains
itself. First we attempt a _nofault copy, if that fails, we try plain.

While using strncpy_from_user_nofault was considered, this wont do, this will
only copy up to the first \0.

for LKM, we just need sys_execve for the filename. same method works as-is.
using lsm_hook_hack wasnt checked, but compatibility via sys_execve kprobe is kept.

devlog:
https://github.com/ximi-libra-test/android_kernel_xiaomi_libra/compare/16e5dce9e7e...16c1f5f521a
https://github.com/backslashxx/mojito_krenol/compare/28642e60d7b...728de0c571c

References:
https://elixir.bootlin.com/linux/v4.14.1/source/include/linux/mm_types.h#L429
https://elixir.bootlin.com/linux/v4.14.1/source/include/linux/lsm_hooks.h

```

### DIFF
--- a/kernel/ksud.c
+++ b/kernel/ksud.c
@@ -175,7 +175,6 @@ first_app_process:
 
 int ksu_handle_pre_ksud(const char *filename)
 {
-
 	if (likely(!ksu_execveat_hook))
 		return 0;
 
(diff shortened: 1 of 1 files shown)


backslashxx:
context:
[`kernel: ksud, core_hook: migrate ksud execution to security_bprm_check`](https://github.com/backslashxx/KernelSU/commit/972ee63c928f71aa61fc806dc133d8b74e2b4b7b)

logs on lkm mode (via sys_execve)
[ksu_lkm_log.log](https://github.com/user-attachments/files/20913797/ksu_lkm_log.log)
from @ukriu 

[dmesg_6.1.txt](https://github.com/user-attachments/files/20924052/dmesg_6.1.txt) from fergus
[logs2electricboogaloo.log](https://github.com/user-attachments/files/20924219/logs2electricboogaloo.log) from @ukriu again
logs for https://github.com/tiann/KernelSU/commit/ab23da81aab782ffa53b392d54639311cf94bdf5


LKM demo
https://github.com/backslashxx/KernelSU/actions/runs/15898080119

LSM / GKI demo prolly unneeded

--

I havent tested lsm hook hack but this can be hooked on different places

5.8+ **bprm_creds_for_exec** + bprm_creds_from_file?
2.6.29 - 5.7 bprm_set_creds (tested)
2.6.0 - 6.15 bprm_check_security (tested)

however for my purposes, I chose bprm_check_security since
1. max compatibility
2. bprm->buf is filled so https://github.com/backslashxx/KernelSU/commit/ca22487e13b0b36ce24e242cd54716799fc31551

PR #2659, opened 2025-06-27
Make "App profile" translatable

igormiguell:
The "App profile" string is now translatable. I don't think this will cause any problems.


tiann:
We don't want this to be translatable; it is a term like "SELinux."

igormiguell:
I noticed that in some languages, like Russian, Japanese, Polish, and others, the translation was done in the local language, so I thought it would be fine.

PR #2662, opened 2025-06-29
Support building for kernel 6.14+ 

hmtheboy154:
Require [this PR](https://github.com/tiann/KernelSU/pull/1785) which is made for 6.8+ to be able to build.

### DIFF
--- a/kernel/selinux/selinux.c
+++ b/kernel/selinux/selinux.c
@@ -86,15 +86,28 @@ static inline u32 current_sid(void)
 
 bool is_ksu_domain()
 {
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 14, 0)
+	struct lsm_context ctx;
+#else
 	char *domain;
 	u32 seclen;
+#endif
 	bool result;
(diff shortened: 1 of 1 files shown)


backslashxx:
wonder if linux or aosp will backport this to 6.12
abi breaking though so not likely
but then again, no device is on 6.12 yet so they can break abi early

hmtheboy154:
> wonder if linux or aosp will backport this to 6.12 abi breaking though so not likely but then again, no device is on 6.12 yet so they can break abi early

yea I saw they backport network changes before, which breaks some of the old realtek wifi drivers that I'm using.

tiann:
Could you address the conflicts? This is necessary for kernels 6.14 and above. Thank you.

hmtheboy154:
sorry I just noticed. Will do

hmtheboy154:
Rebased. I can't test right now due to some issue with my build server. Anyone please check

tiann:
I will check it soon

tiann (approved):

PR #2673, opened 2025-07-09
kernel: added new prctl CMD_GET_MANAGER_UID to get the uid of the crowned manager

rifsxd:
This is useful for root process to get current manager uid.

- Aviraxp


kernel/core_hook.c
```diff
@@ -491,6 +491,17 @@ int ksu_handle_prctl(int option, unsigned long arg2, unsigned long arg3,
 		return 0;
 	}
 
+	if (arg2 == CMD_GET_MANAGER_UID) {
```
5ec1cff: It seems that only manager can get manager uid.

rifsxd: No

rifsxd: All root process can too

rifsxd: Already tested and working with Rezygisk

rifsxd: This feature is needed for upcoming Zygisk-Next too

aviraxp: https://github.com/rifsxd/KernelSU/blob/fa1cac02626283cd38ff4c7b27f8715dcce0ea6d/kernel/core_hook.c#L417

Unlike ksu next, ksu returns early here if uid is not manager

rifsxd: oh damn you are absolutly right i didnt think of that and put it in the very end unlike what i did in next, plus you did the same on your PR too, tho good catch i will move it up before https://github.com/rifsxd/KernelSU/blob/fa1cac02626283cd38ff4c7b27f8715dcce0ea6d/kernel/core_hook.c#L417

PR #2677, opened 2025-07-10
kernel: Introduce CMD_GET_MANAGER_UID prctl option

aviraxp:
This is useful for root process to get current manager uid.


rifsxd:
duplicate of #2673

PR #2696, opened 2025-08-04
Handle unmount for isolated process correctly

aviraxp:
Isolated processes can be directly forked from zygote, but current code doesn't handle it well. Fix it by unmounting unconditionally if isolated process is forked from zygote.

### DIFF
--- a/kernel/core_hook.c
+++ b/kernel/core_hook.c
@@ -62,7 +62,7 @@ static inline bool is_allow_su()
 	return ksu_is_allow_uid(current_uid().val);
 }
 
-static inline bool is_unsupported_uid(uid_t uid)
+static inline bool is_unsupported_app_uid(uid_t uid)
 {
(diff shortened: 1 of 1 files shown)


aviraxp:
I think this cause trouble for some modules. Unlike magisk, ksu doesn't have a userspace daemon to match isolated process with app.

HSSkyBoy:
[I referred to your code](https://github.com/SukiSU-Ultra/SukiSU-Ultra/pull/333), I think there's nothing major about it 💕

aviraxp:
I reconsidered this. Isolated process which directly forks from zygote is just like normal app, so there is no reason apps won't crash but Isolated process will. Also zygote reopens fd before actually fork, so it should be fine.

aviraxp:
I don't use any module that mounts, so maybe feedback is needed for ones especially uses overlay, font modules.

PR #2708, opened 2025-08-15
Reset seccomp filter count when escaping to root

aviraxp:


mio-19:
It doesn't compile on legacy kernel. A new ifdef condition might be needed for legacy kernel. A workaround is  https://github.com/selfmusing/kernel_xiaomi_violet/commit/9596554cfbdab57682a430c15ca64c691d404152

aviraxp:
Legacy kernel doesn't have this issue, why do you pick it

mio-19:
> why do you pick it

Because it is in the main branch

PR #2731, opened 2025-08-30
manager: Optimized import, optimized all libsu shell calls, and fixed WebUI memory leaks

YumeYuka:


tiann:
Thank you for your contribution. Please split it into separate PRs.

PR #2757, opened 2025-09-11
kernel: Add the real UID by parsing the UID from the /data/user_de/0

ShirkNeko:
Prioritize retrieving the application UID from /data/user_de. If this fails, fall back to retrieving it from packages.list.

Fix unstable application UID acquisition

### DIFF
--- a/kernel/throne_tracker.c
+++ b/kernel/throne_tracker.c
@@ -19,7 +19,7 @@ uid_t ksu_manager_uid = KSU_INVALID_UID;
 
 #define USER_DATA_BASE_PATH "/data/user_de"
 #define MAX_SUPPORTED_USERS 32 // Supports up to 32 users
-#define USER_DATA_PATH_LEN 384 // 384 is enough for /data/user_de/{userid}/<package>
+#define DATA_PATH_LEN 384 // 384 is enough for /data/app/<package>/base.apk and /data/user_de/{userid}/<package>
 
 struct uid_data {
 	struct list_head list;
@@ -27,12 +27,58 @@ struct uid_data {
 	char package[KSU_MAX_PACKAGE_NAME];
 };
 
-struct user_scan_context {
+struct user_scan_ctx {
 	struct list_head *uid_list;
 	uid_t user_id;
-	size_t packages_found;
-	size_t errors_count;
+	size_t pkg_count;
+	size_t error_count;
 };
+
+struct user_dir_ctx {
+	struct dir_context ctx;
+	struct user_scan_ctx *scan_ctx;
+};
+
+struct user_id_ctx {
+	struct dir_context ctx;
+	uid_t *user_ids;
+	size_t count;
+	size_t max_count;
+};
+
+struct data_path {
+	char dirpath[DATA_PATH_LEN];
+	int depth;
+	struct list_head list;
+};
+
+struct apk_path_hash {
+	unsigned int hash;
+	bool exists;
+	struct list_head list;
+};
+
+static struct list_head apk_path_hash_list;
+
+struct my_dir_context {
+	struct dir_context ctx;
+	struct list_head *data_path_list;
+	char *parent_dir;
+	void *private_data;
+	int depth;
+	int *stop;
+};
+// https://docs.kernel.org/filesystems/porting.html
+// filldir_t (readdir callbacks) calling conventions have changed. Instead of returning 0 or -E... it returns bool now. false means "no more" (as -E... used to) and true - "keep going" (as 0 in old calling conventions). Rationale: callers never looked at specific -E... values anyway. -> iterate_shared() instances require no changes at all, all filldir_t ones in the tree converted.
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 1, 0)
+#define FILLDIR_RETURN_TYPE bool
+#define FILLDIR_ACTOR_CONTINUE true
+#define FILLDIR_ACTOR_STOP false
+#else
+#define FILLDIR_RETURN_TYPE int
+#define FILLDIR_ACTOR_CONTINUE 0
+#define FILLDIR_ACTOR_STOP -EINVAL
+#endif
 
 static int get_pkg_from_apk_path(char *pkg, const char *path)
 {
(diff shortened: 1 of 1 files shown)


ShirkNeko:
提交 https://github.com/tiann/KernelSU/pull/2757/commits/e47775a7814e913b16094c3c2e8c5e3b05cc865d 后不再回退到packages.list，且支持多用户扫描，并优化 track_throne 扫描触发条件

kernel/core_hook.c
```diff
@@ -203,25 +203,18 @@ int ksu_handle_rename(struct dentry *old_dentry, struct dentry *new_dentry)
...
-	char path[128];
-	char *buf = dentry_path_raw(new_dentry, path, sizeof(path));
-	if (IS_ERR(buf)) {
-		pr_err("dentry_path_raw failed.\n");
-		return 0;
-	}
-
-	if (!strstr(buf, "/system/packages.list")) {
+	//  It still monitors changes to certain system files to trigger scans, but does not rely on packages.list.
+	if (strstr(new_dentry->d_iname, "packages") &&
```
aviraxp: The logic here seems fragile. What if there is a file whose name is pacakge_123.list?

kernel/throne_tracker.c
```diff
@@ -15,14 +17,25 @@
 
 uid_t ksu_manager_uid = KSU_INVALID_UID;
 
-#define SYSTEM_PACKAGES_LIST_PATH "/data/system/packages.list.tmp"
+#define USER_DATA_BASE_PATH "/data/user_de"
+#define USER_DATA_LEGACY_BASE_PATH "/data/user"
```
aviraxp: No need for it. It is "/data/user_de since Android 7.

kernel/throne_tracker.c
```diff
@@ -128,6 +141,231 @@ struct my_dir_context {
...
+	if (IS_ERR(dir_file)) {
+		pr_warn("Failed to open %s: %ld, trying legacy path\n", 
+			USER_DATA_BASE_PATH, PTR_ERR(dir_file));
+		
+		// Try the legacy path
+		dir_file = ksu_filp_open_compat(USER_DATA_LEGACY_BASE_PATH, O_RDONLY, 0);
+		if (IS_ERR(dir_file)) {
+			pr_err("Failed to open both user data paths: %ld\n", PTR_ERR(dir_file));
+			return PTR_ERR(dir_file);
+		}
```
aviraxp: See above.

kernel/throne_tracker.c
```diff
@@ -128,6 +141,231 @@ struct my_dir_context {
...
+	}
+
+	struct path path;
+	for (uid_t user_id = MIN_ANDROID_USER_ID; user_id <= MAX_ANDROID_USER_ID && count < max_users; user_id++) {
```
aviraxp: Not sure if we want to use iterate_dir, but this is fine as is.

kernel/throne_tracker.c
```diff
@@ -15,14 +17,26 @@
 
 uid_t ksu_manager_uid = KSU_INVALID_UID;
 
-#define SYSTEM_PACKAGES_LIST_PATH "/data/system/packages.list.tmp"
+#define USER_DATA_BASE_PATH "/data/user_de"
+#define USER_DATA_LEGACY_BASE_PATH "/data/user"
+#define USER_DATA_PATH_LEN 256 // 256 is enough for /data/user_de/{userid}/<package>
```
aviraxp: It is same as KSU_MAX_PACKAGE_NAME (which should be a bit larger), not sure why it is designed like this.

ShirkNeko: 一般来说256已经足够

ShirkNeko: 它不像/data/app一样拥有更深的目录

aviraxp: 并不足够，app名称最大可以有255个字符

ShirkNeko: 与DATA_PATH_LEN一致？会不会有点太大了

aviraxp:
Ptal, it is fine with me now.

aviraxp:
Do we need additional selinux rules?

tiann:
Do we need to scan all packages under all userIds? 
If userId = 999, the app's uid in the directory is 100000 * 999 + (userId=0's uid). 

If we only need to scan the files in the `/data/user_de/0` directory, we can directly use iterate_dir without recursion, This is much simpler.

ShirkNeko:
> Do we need to scan all packages under all userIds? If userId = 999, the app's uid in the directory is 100000 * 999 + (userId=0's uid).
> 
> If we only need to scan the files in the `/data/user_de/0` directory, we can directly use iterate_dir without recursion, This is much simpler.

Great idea!

ShirkNeko:
This eliminates the overhead caused by unnecessary recursion.

aviraxp:
adb install --user 10 MyApp.apk will directly install app to user 10, not 0. So i think it is a must to scan all dir.

ShirkNeko:
However, in most cases, user 0 is always used as the primary user, and support for multiple users is more of a design reserve.

ShirkNeko:
Filtering and identification using active users

tiann:
If supporting multiple users is indeed necessary, it can be done as follows:  
1. Scan `/data/user_de/0` to get all app UIDs.  
2. (Optional) Scan `/data/user_de` to get all users.  
3. (Optional) Use the same step 1 to get apps installed separately by other users.  

Given that current multi-user environments cannot be completely isolated, I personally believe that installing apps separately in multiple users to evade detection is unnecessary.

ShirkNeko:
> If supporting multiple users is indeed necessary, it can be done as follows:
> 
> 1. Scan `/data/user_de/0` to get all app UIDs.
> 2. (Optional) Scan `/data/user_de` to get all users.
> 3. (Optional) Use the same step 1 to get apps installed separately by other users.
> 
> Given that current multi-user environments cannot be completely isolated, I personally believe that installing apps separately in multiple users to evade detection is unnecessary.

好的，也许我还能改进下

PR #2776, opened 2025-09-25
Revert "Unmount isolated process which forks from zygote unconditionally"

yujincheng08:
Reverts tiann/KernelSU#2747


backslashxx:
make up your mind

merge - revert - reapply - revert

PR #2807, opened 2025-10-17
kernel: core_hook: skip private space recrowning

backslashxx:


backslashxx:
this is actually one of the culprits on that prctl detection that zimperium does.
this is probably enough for 5.10+ devices. (since they are modern ones and fast anyway)

detection is done by comparing a 0xdeadbeef call to a non-0xdeadbeef after calling near misses / branch pre-training.

aviraxp:
I think it is fine. Actually after moving package parsing to /data/user, we can remove the speical hack.

backslashxx:
on mine I actually had to precheck is_allow_su but this is due to me supporting ultra legacy
these garbages are slow that I can see the time differences easy.

[kernel: core_hook: harden prctl handler](https://github.com/backslashxx/KernelSU/commit/e48191fc93f5d37e6290b024f0a50857dcadf0f2)

tiann:
How about removing lines 265 to 270?

backslashxx:
well, you will lose 'manager on private space' feature, atleast with the current throne tracker

backslashxx:
well, you will lose 'manager on private space' feature, atleast with the current throne tracker

the actual issue is on here
https://github.com/tiann/KernelSU/blob/1364710021ee7fa25785526d1ef335f3a4d6ddc0/kernel/core_hook.c#L256-L257

this creates a fast path + slow path

you call it with 0xdeadbeef and its slower.

but this is not noticeable on newer devices
what aggravates is though is those three uid checks and division behind it before it hits !from_root and !from_manager check

for full mitigation something like this

```C
int ksu_handle_prctl(int option, unsigned long arg2, unsigned long arg3,
		     unsigned long arg4, unsigned long arg5)
{
	// if success, we modify the arg5 as result!
	u32 *result = (u32 *)arg5;
	u32 reply_ok = KERNEL_SU_OPTION;
	uid_t current_uid_val = current_uid().val;

	if (current_uid_val < 100000)
		goto skip_multiuser;

	// TODO: find it in throne tracker!
	uid_t manager_uid = ksu_get_manager_uid();
	if (current_uid_val != manager_uid &&
	    current_uid_val % 100000 == manager_uid) {
		ksu_set_manager_uid(current_uid_val);
	}

skip_multiuser:
	bool from_root = !current_uid_val;
	bool from_manager = is_manager();

	barrier();
	if (!from_root && !from_manager) {
		// only root or manager can access this interface
		return 0;
	}
	
	if (KERNEL_SU_OPTION != option) {
		return 0;
	}
```

you keep private space support and gate magic check behind uid check

PR #2818, opened 2025-10-19
Fix kernel panics caused by thread info flag corruption

kerneltoast:


aviraxp (approved):

tiann:
~~5.10 doesn't have this header. The key should be `clear_bit` to atomically clear the flag. We can just use the `clear_bit`.~~

kerneltoast:
5.10 absolutely has `clear_thread_flag()`. It is a very old function.

See: https://elixir.bootlin.com/linux/v5.10/A/ident/clear_thread_flag

tiann:
Sorry, I misread that

PR #2833, opened 2025-10-28
chore: Remove accidentally uploaded .vscode directory and update .gitignore

Prslc:
I noticed that the `.vscode` directory was accidentally uploaded in commit cf361a7, so I’ve removed it and updated the `.gitignore` to prevent it from being accidentally committed again. 

This commit may require a rebase.


Prslc:
Sorry, I didn't notice that `kernel/Makefile` used `.vscode`, I will close this PR

Link: https://github.com/tiann/KernelSU/blob/main/kernel%2FMakefile#L77

PR #2861, opened 2025-11-04
kernel: supercall: allow escalation on ioctl interface

backslashxx:


backslashxx:
related: https://github.com/backslashxx/kernelnosu/commits/update

I actually was trying to do something more before the disablign was added on ksud
https://github.com/backslashxx/kernelnosu/commits/new-supercall-dev

thanks for having it on ksud :+1:

tiann:
Why is this needed?

backslashxx:
for weird calls (e.g. **cd folder ; ./su**, appease execveat? and compat_execve calls like on kali nethunter)
and for x86 that has that syscall hardening. (syscall hooks sometimes fail)

and ofcourse, you can't beat real file for compatibility.

as for most people that use this, to hide delayed syscall.
they just unmount the su binary anyway so its like having magisk at this point.

Ylarod:
Anyone can gain root if no seccomp applied, like `shell`. This breaks the security model

tiann:
If the syscall hook fails, we recommend using `ksud debug su`. Alternatively, you can call `SYS_reboot` along with `ioctl` to gain root access. Allowing anyone to grant root access through ioctl is not advisable.

backslashxx:
@tiann 

> Alternatively, you can call SYS_reboot along with ioctl to gain root access

<details>
<summary> img </summary>

image

</details>

as you see that is what I am doing.

grab an fd from sys_reboot, escalate via ioctl
but then KSU_IOCTL_GRANT_ROOT is only for root and manager.

```C
#include <sys/syscall.h>
#include <unistd.h>
#include <termios.h>
#include <string.h>
#include <sys/ioctl.h> 
#include <fcntl.h>

// zig cc -target arm-linux -s -Os su.c -static -Wl,--gc-sections
// /tmp/optane/openwrt-mr7350/staging_dir/host/bin/sstrip a.out

static int denied(void)
{
	const char *error = "Denied\n";
	syscall(SYS_write, 2, error, strlen(error));
	return 1;
}

// ksu's new supercall
#define KSU_INSTALL_MAGIC1 0xDEADBEEF
#define KSU_INSTALL_MAGIC2 0xCAFEBABE
#define KSU_IOCTL_GRANT_ROOT _IOC(_IOC_NONE, 'K', 1, 0)

int main(int argc, const char **argv, const char **envp)
{
	int fd = 0;
	
	syscall(SYS_reboot, KSU_INSTALL_MAGIC1, KSU_INSTALL_MAGIC2, 0, (void *)&fd);
	if (fd == 0) {
		return denied();
	} else {
		int ret = syscall(SYS_ioctl, fd, KSU_IOCTL_GRANT_ROOT, 0);
		syscall(SYS_close, fd); // close it here regardless
		if (ret < 0)
			return denied();
	}

	argv[0] = "su";

	syscall(SYS_execve, "/data/adb/ksud", argv, envp);
	
	return denied();
}
```

https://github.com/backslashxx/kernelnosu/blob/e0671194212a2b96e3da96dd3525eaab4d856c31/src/su.c

but oh well, its fine ig,

backslashxx:
@Ylarod 
> Anyone can gain root if no seccomp applied, like `shell`. This breaks the security model

so add another perm_check? where check only for allowed uid?

PR #2862, opened 2025-11-04
kernel: supercall: allow escalation on ioctl interface

backslashxx:


backslashxx:
take 2

this addresses what @Ylarod said

image


ref: https://github.com/tiann/KernelSU/pull/2861#issuecomment-3486265257

usage example: https://godbolt.org/z/Kh91EG5eq

hmtheboy154:
+1 

If this is safe to merge and can fix requiring `syscall_hardening=off` on x86 here, then LGTM.

kernel/supercalls.c
```diff
@@ -47,6 +47,12 @@ bool always_allow(void)
...
 
+bool allowed_for_su(void)
+{
+    bool is_allowed = is_manager() || ksu_is_allow_uid(current_uid().val);
+    return is_allowed;
+}
+
 static int do_grant_root(void __user *arg)
 {
     // Check if current UID is allowed
```
Ylarod: LGTM, remove dup check in do_grant_root, sorry for forgetting there is a check already
```
    // Check if current UID is allowed
    bool is_allowed = is_manager() || ksu_is_allow_uid(current_uid().val);
    if (!is_allowed) {
        return -EPERM;
    }
```

Ylarod (changes requested):

kernel/supercalls.c
backslashxx: resolved. Thanks for giving a second chance.

PR #2869, opened 2025-11-05
support mainline kernel

aaaaaaaa-815:


aaaaaaaa-815:
based on https://github.com/tiann/KernelSU/pull/2662

PR #2895, opened 2025-11-09
manager: if manager incompatible with current kernel, don't save ksud

AlexLiuDev233:
I think we should'nt install kernelsu's userspace when manager incompatible with current kernel
this maybe cause a lot of bug, for example, when user install 2.x kernelsu manager, but not update 
his kernel to 2.x, ksud will return "kernel version 0", because ksud incompatible with old supercall impl

### DIFF
--- a/kernel/allowlist.c
+++ b/kernel/allowlist.c
@@ -397,6 +397,7 @@ close_file:
     filp_close(fp, 0);
 unlock:
     mutex_unlock(&allowlist_mutex);
+    kfree(_cb);
 }
 
 void persistent_allow_list()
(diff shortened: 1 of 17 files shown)


tiann:
This will cause lots of bug too.
If user uses manager 2.0 and then upgrade kernel in manager, and then reboot, the ksud is the old one and cannot communicate with the kernel.

AlexLiuDev233:
> This will cause lots of bug too. If user uses manager 2.0 and then upgrade kernel in manager, and then reboot, the ksud is the old one and cannot communicate with the kernel.

hmmm, maybe i need add a ksud update when user update kernel to solve this problem? i will write a new commit

AlexLiuDev233:
@tiann can you review again?

tiann (approved):

aviraxp (approved):

PR #2900, opened 2025-11-09
kernel: Migrate build logic from Makefile to Kbuild

dabao1955:
This commit replaces the legacy kernel/Makefile with a dedicated Kbuild file to align with upstream kernel build conventions. The new kernel/Kbuild defines all KernelSU object files, SELinux integration paths, and versioning logic, including:

- Git-based version detection for both in-tree (submodule) and out-of-tree builds
- Fallback to default version (16) when .git is absent
- Manager package name, signature size, and hash configuration via build args
- Proper include paths for SELinux headers and errno definitions

The old Makefile content was largely duplicated build logic that belongs in Kbuild. Only the essential compiler flags (e.g., -Wno-*) are retained in the Makefile for compatibility. This change improves maintainability and follows standard Linux kernel module build practices.

### DIFF
--- a/kernel/Kbuild
+++ b/kernel/Kbuild
@@ -1,0 +1,78 @@
+kernelsu-objs := ksu.o
+kernelsu-objs += allowlist.o
+kernelsu-objs += app_profile.o
+kernelsu-objs += apk_sign.o
+kernelsu-objs += sucompat.o
+kernelsu-objs += syscall_hook_manager.o
+kernelsu-objs += throne_tracker.o
+kernelsu-objs += pkg_observer.o
+kernelsu-objs += setuid_hook.o
+kernelsu-objs += kernel_umount.o
+kernelsu-objs += supercalls.o
+kernelsu-objs += feature.o
+kernelsu-objs += ksud.o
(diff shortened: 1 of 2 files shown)


Aquarius223:
KernelSU has updated its version calculation method in [**v2.0.0**](https://github.com/tiann/KernelSU/commit/1565f4ba8), so could the version calculation instructions [**here**](https://github.com/tiann/KernelSU/pull/2900/commits/bc0e7fef7dc315e50efb032d9e812c821b4236b8#diff-0faa202c61d64b62f628d96762c5dc4a5deda4020c23f4fa9475d89e40e33f49R43) also be updated in this update?

dabao1955:
> KernelSU has updated its version calculation method in [**v2.0.0**](https://github.com/tiann/KernelSU/commit/1565f4ba8), so could the version calculation instructions [**here**](https://github.com/tiann/KernelSU/pull/2900/commits/bc0e7fef7dc315e50efb032d9e812c821b4236b8#diff-0faa202c61d64b62f628d96762c5dc4a5deda4020c23f4fa9475d89e40e33f49R43) also be updated in this update?

This change based on main branch

PR #2912, opened 2025-11-11
kernel: Fix task flag marking for root and shell UID

aviraxp:
One case is calling su with su.

### DIFF
--- a/kernel/setuid_hook.c
+++ b/kernel/setuid_hook.c
@@ -149,8 +149,14 @@ int ksu_handle_setresuid(uid_t ruid, uid_t euid, uid_t suid)
     }
 
     // for init forks adbd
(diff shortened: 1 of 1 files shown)


kernel/setuid_hook.c
```diff
@@ -151,6 +148,17 @@ int ksu_handle_setresuid(uid_t ruid, uid_t euid, uid_t suid)
         ksu_clear_task_tracepoint_flag(current);
     }
 
+    // for init forks adbd
+    {
+        struct task_struct *parent;
+        rcu_read_lock();
+        parent = rcu_dereference(current->real_parent);
+        if (parent && parent->pid == 1 && new_uid == 2000) {
```
tiann: You can rcu_read_unlock directly after obtaining the ppid, and then clear flag.

PR #2918, opened 2025-11-11
Improve compatibility with Samsung 6.6 GKI 2.0 devices

Fede2782:
This PR aims to fix the two main issues Samsung GKI devices encounter when using GKI 2.0 images. This is for Samsung devices which really support GKI images as some devices, despite being on 5.10+, don't support GKI images as the OEM and the SoC Manufacturer include essential drivers in the kernel image instead of shipping them as modules.

One patch is included.

1) min_KDP support: Samsung devices with 6.6 kernel ship a Bluetooth driver (bluetooth.ko) which depends on these three KDP functions. Import a minimal KDP driver which behaves like the normal KDP driver when KDP is disabled by calling the required kernel functions.

Despite Samsung always adding a lot of obstacles this might really improve the situation as building kernels from Samsung sources usually takes time due to Samsung OSRC taking a lot of time to release them publicly.

Alternatively to this, a GKI variant specifically for Samsung KMIs (android12-5.10, android13-5.15, android14-6.1 and android15-6.6) can be made like is being done with ARCVM/WSA.

### DIFF
--- a/.github/patches/6.6/add-min_kdp-symbols-2024-09.patch
+++ b/.github/patches/6.6/add-min_kdp-symbols-2024-09.patch
@@ -1,0 +1,86 @@
+From acb31159fa5b4a4eb9197624e15fbf6e51802b8c Mon Sep 17 00:00:00 2001
+From: Fede2782 <78815152+Fede2782@users.noreply.github.com>
+Date: Wed, 23 Jul 2025 18:35:13 +0200
+Subject: [PATCH] add min_kdp symbols
+
+This adds symbols provided by min_kdp which are required for bluetooth driver
+---
+ android/abi_gki_aarch64.stg | 41 +++++++++++++++++++++++++++++++++++++
+ 1 file changed, 41 insertions(+)
+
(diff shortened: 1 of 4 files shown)


backslashxx:
hah, we need to shim symbols now. Samsung "GKI" is a mess

rifsxd:
wdym "Samsung 6.6 GKI" i wont ever consider that as GKI with the amount of broken shit they add to the linux kernel

Fede2782:
> wdym "Samsung 6.6 GKI" i wont ever consider that as GKI with the amount of broken shit they add to the linux kernel

Mediatek and Exynos Samsung devices are fully GKI, it's a fact. It's just this issue for 6.6 and sdfat for 5.10-6.1. I don't know about Qualcomm but this is way better than most other OEMs (Google excluded). With this patch KernelSU GKI would have worked without any issue on a lot of devices like S25, A34, A14, M53, A56, A36 and probably more with 6.6. However GKI mode is depreciated so closing this.

PR #2922, opened 2025-11-12
kernel: Prune allowlist only after boot completed

aviraxp:
For unknown reason, packages.list is not reliable during boot for oplus devices, so we have to disable pruning and re-run pruning after boot.

### DIFF
--- a/kernel/throne_tracker.c
+++ b/kernel/throne_tracker.c
@@ -300,7 +300,7 @@ static bool is_uid_exist(uid_t uid, char *package, void *data)
     return exist;
 }
 
-void track_throne()
+void track_throne(bool prune_only)
 {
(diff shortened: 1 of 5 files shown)


kernel/throne_tracker.c
```diff
@@ -357,27 +357,29 @@ void track_throne()
...
-    // first, check if manager_uid exist!
-    bool manager_exist = false;
-    list_for_each_entry (np, &uid_list, list) {
-        // if manager is installed in work profile, the uid in packages.list is still equals main profile
-        // don't delete it in this case!
-        int manager_uid = ksu_get_manager_uid() % 100000;
-        if (np->uid == manager_uid) {
-            manager_exist = true;
-            break;
+    if (!prune_only) {
```
backslashxx: maybe goto
if (prune_only) goto prune

PR #2929, opened 2025-11-13
support metamodule, remove built-in overlayfs mount

Ylarod:


backslashxx:
I don't like that you killed A/B updates.
Everything else seems fine though.

but sure I guess you can enforce that theres no more /data/adb/modules_update

tiann:
If you need, feel free to open a PR.

backslashxx:
yeah right, I guess I can't really say shit since I can't rust.

tiann:
When we use overlayfs, all the data of the modules is packaged inside the ext4 image, and the data is consistent; however, once the modules are placed in a directory, this advantage is lost, making AB updates less natural. There may be good solutions, but we haven't thought of them yet. Our priority is to advance the metamodule first.

backslashxx:
metamodule can handle A/B but we have to be sourced on top, right before installer.sh.
then we can handle mv -f modules_update/ modules/ on metamount

tiann:
You can use `metainstall.sh`, complete your tasks, and then call `install_module`.

PR #2952, opened 2025-11-18
manager: add inset support to webui

KOWX712:
ref: https://github.com/MMRLApp/WebUI-X-Portable/blob/master/webui/src/main/kotlin/com/dergoogler/mmrl/webui/model/Insets.kt

Co-Authored-By: Der_Googler <54764558+dergoogler@users.noreply.github.com>
Signed-off-by: KOWX712 <leecc0503@gmail.com>


KOWX712:
This standard is used in WebUI-X (also known as MMRL) for quite a while, many module's webui has already added support to this, here is a simple guide to add inset support in webui:
```css
@import url('https://mui.kernelsu.org/internal/insets.css');

body {
    padding-top: var(--window-inset-top, 0px);
    padding-bottom: var(--window-inset-bottom, 0px);
}
```

Old WebUI vs New WebUI
<p>
  
  
</p>

PR #2972, opened 2025-11-23
kernel: Unmount all isolated process which forks from zygote

aviraxp:


tiann (approved):

backslashxx:
image

so I guess at this point either you will have to manage kernel umount (managedFeatures) in zn or pre-delete the entries that zn will umount.

PR #2999, opened 2025-11-28
support kernel 6.18

aaaaaaaa-815:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/diff/security/selinux/hooks.c?id=v6.18-rc7&id2=v6.18-rc6


hmtheboy154:
I think you should change the commit to "support kernel 6.18"

Ylarod:
just use a same var name, a lot of dup code

aaaaaaaa-815:
> just use a same var name, a lot of dup code

got it

PR #3005, opened 2025-11-29
add hasFragileUserData to keep manager data when uninstall

u9521:
现在管理器有一些设置存在SharedPreferences中，例如
- 主题
- webview调试
- 超级用户是否显示系统应用
- 检查更新
...

不确定以后会不会多起来。添加hasFragileUserData方便选择卸载后保留管理器相关的数据

方便以下操作：

- 卸载管理器避免检测
- 保留数据从ci降级到稳定版(~尽管不推荐，但还是能做到~)

### DIFF
--- a/manager/app/src/main/AndroidManifest.xml
+++ b/manager/app/src/main/AndroidManifest.xml
@@ -10,6 +10,7 @@
         android:dataExtractionRules="@xml/data_extraction_rules"
         android:enableOnBackInvokedCallback="false"
         android:fullBackupContent="@xml/backup_rules"
+        android:hasFragileUserData="true"
         android:icon="@mipmap/ic_launcher"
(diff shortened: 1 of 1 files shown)


aviraxp:
Note hasFragileUserData is broken on A11-, making app cannot get uninstalled. But it is fine for us.

u9521:
> Note hasFragileUserData is broken on A11-, making app cannot get uninstalled. But it is fine for us.

I googled this issue and found [this](https://gist.github.com/agnostic-apollo/58def38fdb1c90563454974108ae52df). Triggering this bug is quite specific: it must be on **Android 10** with the external SD card set as a removable device. However, we have already dropped support for non-GKI devices, and I think users use non-GKI devices encountering this bug would use ADB to uninstall the manager.

PR #3012, opened 2025-11-29
manager: fix webui load indefenitely if app list is not initialized

KOWX712:
this can happen when we launch webui activity directly with shortcut, or you can simply replicated it by running `am start -n "me.weishu.kernelsu/.ui.webui.WebUIActivity" -e id zygisksu -e name zn`

Signed-off-by: KOWX712 <leecc0503@gmail.com>


sanasaiyan-rgb:
> this can happen when we launch webui activity directly with shortcut, or you can simply replicated it by running `am start -n "me.weishu.kernelsu/.ui.webui.WebUIActivity" -e id zygisksu -e name zn`

PR #3018, opened 2025-11-30
Restore automatic builds for GKI

VeryBaaad:
For some people, GKI is a very important product, and none of us want it to be lost. If you don't want to maintain GKI, you should only show a prompt, rather than completely disabling KernelSU's official build of GKI.
At the same time, for projects that rely on GKI (such as 'MagiskOnWSA'), they will also be unable to build, and everyone hopes that it can be downloaded quickly rather than having to build it themselves.

### DIFF
--- a/.github/workflows/build-gki-image.yml
+++ b/.github/workflows/build-gki-image.yml
@@ -1,0 +1,20 @@
+name: Build Android GKI Image
+on:
+  workflow_call:
+
+jobs:
+  build-a12-kernel:
+    uses: ./.github/workflows/build-kernel-a12.yml
+    secrets: inherit
+  build-a13-kernel:
+    uses: ./.github/workflows/build-kernel-a13.yml
+    secrets: inherit
+  build-a14-kernel:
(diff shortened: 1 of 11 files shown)


CanerKaraca23:
Instead of removing, maybe disabling it on release workflow making much more sense, so if people needs GKI, they can just easily fork and run build for themselves.

aviraxp:
If someone wants to maintain GKI, they can have a standalone repo. I see no reason to keep it in repo when no one will guarantee it will work.

VeryBaaad:
> If someone wants to maintain GKI, they can have a standalone repo. I see no reason to keep it in repo when no one will guarantee it will work.

You could do like the person above and disable it, which might be better. I added it because I think it makes building more convenient, rather than having to manually tweak actions scripts or compile the Linux kernel locally (for me, I've never successfully compiled it, and I also can't easily find the corresponding toolchain).

luyanci:
If someone still want to use gki,you can use this repo.

https://github.com/luyancib-org/KSU-GKI-Standalone/

PR #3028, opened 2025-12-02
kernel: fix root_groups defs

aaaaaaaa-815:
https://elixir.bootlin.com/linux/v6.7-rc1/source/include/linux/cred.h

### DIFF
--- a/kernel/app_profile.c
+++ b/kernel/app_profile.c
@@ -13,7 +13,11 @@
 #include "selinux/selinux.h"
 #include "syscall_hook_manager.h"
 
-static struct group_info root_groups = { .usage = ATOMIC_INIT(2) };
+#if LINUX_VERSION_CODE >= KERNEL_VERSION (6, 7, 0)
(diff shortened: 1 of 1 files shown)


Ylarod:
we never use gcc

aaaaaaaa-815:
> we never use gcc

Will actually it's a change in kernel 6.7 but it can compile when using clang

CoelacanthusHex:
IMO
1. It should be intro, since upstream changed the type indeed. And the implementation of refcount_t may change in future.
2. In current kernel and ksu implementations, it didn't violate any rule of C, it uses nest initialization (allow a flatten brace init list to init nest struct).
```c
typedef struct {
	int counter;
} atomic_t;

struct refcount_struct {
	atomic_t refs;
} refcount_t;

// REFCOUNT_INIT(0) do
refcount_t x = { .refs = 0 };

// ATOMIC_INIT(0) do
refcount_t x = { 0 };

// Two forms above is equal.
```
3. Please attach the build output and the GCC version you used. There may be a bug of GCC.

PR #3032, opened 2025-12-03
kernel: use our cred to unmount, remove sepolicy of zygote

5ec1cff:


kernel/kernel_umount.c
braiann401-droid: ### #3042

PR #3040, opened 2025-12-05
kernel: supercalls: allow userspace to pull list entries

backslashxx:


backslashxx:
userspace handling https://godbolt.org/z/MTcqMGG5o

image


<details> 


```C
#include <sys/syscall.h>
#include <sys/ioctl.h>
#include <unistd.h>
#include <stdio.h>
#include <errno.h>
#include <string.h>
#include <stdint.h>
#include <stdlib.h>

#define KSU_INSTALL_MAGIC1 0xDEADBEEF
#define KSU_INSTALL_MAGIC2 0xCAFEBABE

struct ksu_add_try_umount_cmd {
	uint64_t arg; // char ptr, this is the mountpoint
	uint32_t flags; // this is the flag we use for it
	uint8_t mode; // denotes what to do with it 0:wipe_list 1:add_to_list 2:delete_entry
};

#define KSU_UMOUNT_WIPE 0  // ignore everything and wipe list
#define KSU_UMOUNT_ADD 1   // add entry (path + flags)
#define KSU_UMOUNT_DEL 2   // delete entry, strcmp
#define KSU_UMOUNT_GETSIZE 3   // get list size
#define KSU_UMOUNT_GETLIST 4   // get list

#define KSU_IOCTL_ADD_TRY_UMOUNT _IOC(_IOC_WRITE, 'K', 18, 0)

int main(int argc, char *argv[]) 
{
	int fd = 0; // we get that fd here

	syscall(SYS_reboot, KSU_INSTALL_MAGIC1, KSU_INSTALL_MAGIC2, 0, (void *)&fd);

	if (!fd) {
		printf("sys_reboot failed\n");
		return 1;
	}

	printf("[+] fd : %d\n", fd);

	struct ksu_add_try_umount_cmd cmd = {0};
	
	size_t total_size = 0;
	
	cmd.arg = (uint64_t)&total_size;
	cmd.flags = 0;
	cmd.mode = KSU_UMOUNT_GETSIZE;

	int ret = ioctl(fd, KSU_IOCTL_ADD_TRY_UMOUNT, &cmd);
	if (ret < 0) {
		printf("[-] KSU_IOCTL_ADD_TRY_UMOUNT failed\n");
		return 1;
	}

	if (!total_size)
		return 1;
	
	printf("total_size: %zu \n", total_size);

	// now we can prepare the same size of memory	
	
	void *buffer = malloc(total_size);
	if (!buffer)
		return 1;

	memset(buffer, 0, total_size);

	cmd.arg = (uint64_t)buffer;
	cmd.flags = 0;
	cmd.mode = KSU_UMOUNT_GETLIST;

	ret = ioctl(fd, KSU_IOCTL_ADD_TRY_UMOUNT, &cmd);
	if (ret < 0) {
		printf("[-] KSU_IOCTL_ADD_TRY_UMOUNT failed\n");
		return 1;
	}

	// now we pointerwalk
	const char *char_buf = (const char *)buffer;
	do {
		printf("list_entry: %s \n", char_buf);
		char_buf = char_buf + strlen(char_buf) + 1;
	} while (*char_buf);

	return 0;
}
```

</details>

backslashxx:
@aviraxp

aviraxp:
What's this usecase? If it is to solve confliction with other umount methods, this is not enough:

1) Device boots with kernel umount disabled
2) Zygisk umounted stuff (Note, in rezygisk/neozygisk they just umount stuff in zygote, so zygote is clean since first app fork)
3) User turns on kernel umount, conflict again

backslashxx:
I thought youll be deleting the mount entries that youll unmount / not included to clean ns, and with this you can pull mount entries
so you know which to delete

aviraxp:
All zygisk solutions don't umount at new app fork time now, so there will always be timing issue.

PR #3043, opened 2025-12-05
ci: Run rust checks for ksuinit and fix them

aviraxp:


CanerKaraca23:
I suggest also adding `ksuinit` to Dependabot.

CanerKaraca23:
@aviraxp You should add new one to Dependabot, `userspace` will not work alone, you need to add both `userspace/ksud` and `userspace/ksuinit`.

PR #3055, opened 2025-12-07
manager: Add execBool method to WebViewInterface

DerGoogler:
This allows the user to build a `SuFile` like class in JavaScript.

### DIFF
--- a/manager/app/src/main/java/me/weishu/kernelsu/ui/webui/WebViewInterface.kt
+++ b/manager/app/src/main/java/me/weishu/kernelsu/ui/webui/WebViewInterface.kt
@@ -37,7 +37,7 @@ class WebViewInterface(
     }
 
     @JavascriptInterface
-    fun execBool(cmd: String): Boolean {
+    fun execResult(cmd: String): Boolean {
         return withNewRootShell(true) { ShellUtils.fastCmdResult(this, cmd) }
     }
     


manager/app/src/main/java/me/weishu/kernelsu/ui/webui/WebViewInterface.kt
```diff
@@ -36,6 +36,11 @@ class WebViewInterface(
         return withNewRootShell(true) { ShellUtils.fastCmd(this, cmd) }
     }
 
+    @JavascriptInterface
+    fun execBool(cmd: String): Boolean {
+        return withNewRootShell { ShellUtils.fastCmdResult(this, cmd) }
```
aviraxp: Why is this not in global ns?

DerGoogler: Ohhh, sorry. Absolutely forgot about this. 👍

aviraxp:
Sorry, may i know how you will use this api to build a SuFile like class?

DerGoogler:
Yeah, and for people for needs to check if a command was successful

manager/app/src/main/java/me/weishu/kernelsu/ui/webui/WebViewInterface.kt
```diff
@@ -36,6 +36,11 @@ class WebViewInterface(
         return withNewRootShell(true) { ShellUtils.fastCmd(this, cmd) }
     }
 
+    @JavascriptInterface
+    fun execBool(cmd: String): Boolean {
```
aviraxp: execBool -> execResult

tiann:
We already have the following link: https://github.com/tiann/KernelSU/blob/main/manager/app/src/main/java/me/weishu/kernelsu/ui/webui/WebViewInterface.kt#L40

The callback will provide you with `errno`, which is equivalent to `fastCmdResult`.

```javascript
import { exec } from 'kernelsu';

const { errno, stdout, stderr } = await exec('ls -l', { cwd: '/tmp' });
if (errno === 0) {
    // success
    console.log(stdout);
}
```

aviraxp:
> We already have the following link: https://github.com/tiann/KernelSU/blob/main/manager/app/src/main/java/me/weishu/kernelsu/ui/webui/WebViewInterface.kt#L40
> 
> The callback will provide you with `errno`, which is equivalent to `fastCmdResult`.
> 
> ```javascript
> import { exec } from 'kernelsu';
> 
> const { errno, stdout, stderr } = await exec('ls -l', { cwd: '/tmp' });
> if (errno === 0) {
>     // success
>     console.log(stdout);
> }
> ```
> 
Please use this way.

PR #3088, opened 2025-12-21
ksuinit: update dependencies and migrate to 2024

CanerKaraca23:
`goblin` version is too old, i recommend to upstream it since it's fork.
because of it, `scroll` cannot be updated.

Edits welcomed.


aviraxp:
goblin and kernlog can be updated to use upstream version. (kernlog needs to manually point to https://github.com/kstep/kernlog.rs/commit/ff0b1bd6d5261eae0fa297cec6951fe8d982151a)

PR #3095, opened 2025-12-22
ksuinit: use upstream

CanerKaraca23:
https://github.com/tiann/KernelSU/pull/3088#issuecomment-3682702033

https://github.com/Kernel-SU/rustix can be upstreamed for update.


yujincheng08:
Reverted, see https://github.com/bytecodealliance/rustix/pull/1528#issuecomment-3690254613

PR #3107, opened 2025-12-25
kernel: append to init.rc instead of prepending to atrace.rc

5ec1cff:
Even though they are all triggered in post-fs-data stage, init.rc get parsed and triggered first. This will make module scripts run eariler and help prevent race condition between module rc executes and service starts. This also aligns behavior with Magisk.

### DIFF
--- a/kernel/ksud.c
+++ b/kernel/ksud.c
@@ -283,8 +283,7 @@ static ssize_t read_proxy(struct file *file, char __user *buf, size_t count,
     if (ret != 0 || ksu_rc_pos >= ksu_rc_len) {
         return ret;
     } else {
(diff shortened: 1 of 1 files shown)


5ec1cff:
@backslashxx To support Android versions below 11, it might be necessary to inject the `/init.rc` file, just like [Magisk does](https://github.com/topjohnwu/Magisk/blob/01cb75eaefbd14c2d10772ded3942660ebf0285f/native/src/init/rootdir.cpp#L317-L322).

backslashxx:
I will update, but yeah I noticed this earlier so I added atrace as another option.

<details>

image

</details>

all good on A10

<details>
image

</details>

PR #3139, opened 2026-01-05
manager: add file chooser support to WebUI

KOWX712:


KOWX712:
example usecase:
```html
<input type="file" multiple accept=".jpg,.png,">
```
```html
<input type="file" accept="image/*">
```

PR #3154, opened 2026-01-14
manager: remove Coil dependency and add OkHttp directly

u9521:
Removed the Coil Compose dependency as it's no longer needed after migrating to MiuiX for loading app icons.

Added OkHttp dependency directly (previously it was a transitive dependency through Coil).

### DIFF
--- a/manager/gradle/libs.versions.toml
+++ b/manager/gradle/libs.versions.toml
@@ -20,7 +20,6 @@ haze = "1.7.1"
 capsule = "2.1.2"
 okhttp = "5.3.2"
 
-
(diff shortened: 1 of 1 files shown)


manager/gradle/libs.versions.toml
```diff
@@ -19,6 +18,8 @@ cmaker = "1.2"
 miuix = "0.8.0-rc04"
 haze = "1.7.1"
 capsule = "2.1.2"
+okhttp = "5.3.2"
+
```
aviraxp, line 22: redundant line

PR #3189, opened 2026-01-28
kernel: Fix `setup_selinux` using `__task_cred` directly

jsoltan226:
This PR fixes an unsafe direct modification of task credentials in the SELinux integration code used by KernelSU.

In kernel/selinux, KernelSU currently accesses and mutates the struct cred returned by __task_cred(). While this may work on many kernels, it can cause kernel crashes on systems that enforce additional protections on credential memory, such as Samsung devices using the RKP (UH) hypervisor.

On such systems, the credential pages returned by __task_cred() may be write-protected, and writing to them directly can result in a panic during early boot or when launching the KernelSU manager app.

More recent KernelSU versions already address this issue in the “escape to root” logic by switching to the proper kernel APIs (prepare_creds() / commit_creds()), which avoids directly modifying protected credential memory.
However, setup_selinux() (in selinux/selinux.c) still accesses __task_cred() directly. This patch updates that code path to use the same safe credential handling approach, eliminating the remaining unsafe

This change improves compatibility with kernels that enforce credential memory protections (e.g. Samsung kernels with RKP enabled) and removes the need for users to disable CONFIG_UH as a workaround.

I do not currently have access to a GKI-based device to test this on real hardware. However:
The change builds successfully in the AOSP mainline kernel tree
I've backported the logic to my Samsung 4.14 non-GKI kernel tree and KernelSU 0.9.5 works (even with CONFIG_UH=y)

Additional note:
It may be helpful to document this behavior in the non-GKI integration notes, particularly for Samsung kernels that enable RKP/UH.
Several users (myself included) have run into build-time or runtime issues on Samsung kernels due to this interaction and initially worked around it by disabling RKP or switching to a non-kernel-based root solution. Making this information more visible could help others avoid those problems.
I’d be happy to help update or contribute to the relevant documentation if that would be useful.

### DIFF
--- a/kernel/selinux/selinux.c
+++ b/kernel/selinux/selinux.c
@@ -53,7 +53,7 @@ static int transive_to_domain(const char *domain, struct cred *cred)
 
 void setup_selinux(const char *domain, struct cred *cred)
 {
-    if (transive_to_domain(domain, cred) {
+    if (transive_to_domain(domain, cred)) {
         pr_err("transive domain failed.\n");
         return;
     }


jsoltan226:
ehhh third time's a charm I guess
I have no idea how this built in AOSP lol, maybe it wasn't building in my changes?
Whatever...

In newer versions, there appear to be more things that trigger RKP-based crashes, since when I tried to launch a backport of release 3.0.0, I was getting the same
`Unexpected single-step exception at EL1`
kernel log spam and then reset as when I was getting with 0.9.5 without this patch

Also, after patching the commandline to set (among others) `androidboot.verifiedbootstate` to `green` in early kernel startup, even WITH the patch on 0.9.5, RKP (or maybe DEFEX?) starts causing crashes again in the same way as before...
So don't close this PR yet, I'll try to look into the other issues. Also I have set up code completion properly now, so there shouldn't be any silly back and forth like this anymore

aviraxp:
I will merge this commit ahead. It is reasonable.

PR #3205, opened 2026-02-08
ksud: Return the complete error when metamodule script exec failed

Tools-cx-app:
I think that when a meta module's meta mount fails, it should return a full err instead of an exit status, as this makes debugging difficult.

### DIFF
--- a/userspace/ksud/src/metamodule.rs
+++ b/userspace/ksud/src/metamodule.rs
@@ -267,7 +267,7 @@ pub fn exec_mount_script(module_dir: &str) -> Result<()> {
         .envs(crate::module::get_common_script_envs())
         .env("MODULE_DIR", module_dir);
 
-    if fs::exists(defs::METAMODULE_DEBUG)? {
+    if Path::new(defs::METAMODULE_DEBUG).exists() {
         command
             .stdout(File::create(format!(
                 "{}.out.log",


5ec1cff:
It is recommended to redirect stdout and stderr to files (e.g., `/data/adb/ksu/log/metamodule_{mount,uninstall}_{stdout,stderr}.log`), and it would be best to add a switch for this feature, as it is a debugging feature.

Tools-cx-app:
> It is recommended to redirect stdout and stderr to files (e.g., `/data/adb/ksu/log/metamodule_{mount,uninstall}_{stdout,stderr}.log`), and it would be best to add a switch for this feature, as it is a debugging feature.

Completed

userspace/ksud/src/metamodule.rs
```diff
@@ -229,11 +230,18 @@ pub fn exec_metauninstall_script(module_id: &str) -> Result<()> {
...
         .envs(crate::module::get_common_script_envs())
         .env("MODULE_ID", module_id)
-        .status()?;
+        .output()?;
+
+    let err = String::from_utf8_lossy(&result.stderr);
+
+    if fs::exists(defs::METAMODULE_DEBUG)? {
+        fs::write(defs::METAMODULE_METAUNINSTALL_SCRIPT_LOG, err.to_string())?;
+    }
```
5ec1cff:
```suggestion
    let mut command = Command::new(assets::BUSYBOX_PATH);
    command.args(["sh", metauninstall_path.to_str().unwrap()])
        .current_dir(metauninstall_path.parent().unwrap())
        .envs(crate::module::get_common_script_envs())
        .env("MODULE_ID", module_id);
    if fs::exists(defs::METAMODULE_DEBUG)? {
        command.stdout(File::create(...)?)
            .stderr(File::create(...)?);
    }
    let result = command.status()?;
```

Issue #3176, opened 2026-01-21
Incorrect Private Space designation in Superuser menu


0ctobot:
### Please check before submitting an issue

- [x] I have searched the issues and haven't found anything relevant
- [x] I will upload bugreport file in KernelSU Manager - Settings - Report log
- [x] I know how to reproduce the issue which may not be specific to my device

### Describe the bug

When KSU Manager is installed to Private Space, all apps in the Superuser menu are marked with the Private Space badge despite being installed to the primary profile.

This appears to be a visual bug, which has no impact on functionality, but I wanted to bring this to your attention regardless.

### To Reproduce

- 1. Install KSU Manager to Private Space
- 2. Ensure at least one user app is installed to primary profile
- 3. Check Superuser menu
- 4. Observe that user apps on primary profile incorrectly display the Private Space badge 

### Expected behavior

The expectation is that only user apps installed to Private Space would be displayed with the Private Space badge.

### Screenshots

In the following screenshot KSU Manager is currently the only app installed to Private Space, yet all installed apps are marked with the Private Space badge in Superuser menu:

<img width="1080" height="2364" alt="Image" src="https://github.com/user-attachments/assets/a9fb2e4d-c2a1-4cf1-8f73-377104c21978" />

### Logs

I will attach a bug report, because I agreed to do so upon opening this issue, although I don't believe it's relevant to this particular situation.

[KernelSU_bugreport_2026-01-21_18_16.tar.gz](https://github.com/user-attachments/files/24782468/KernelSU_bugreport_2026-01-21_18_16.tar.gz)

### Device info

- Device: Pixel 10 Pro Fold
- OS Version: Android 16 (CP11.251209.007.A1)
- KernelSU Version: Kernel - 32275, Manager - 32288
- Kernel Version: 6.1.121

PR #3236, opened 2026-02-26
Add coil3 + appiconloader

YuKongA:
#3226 #3227


u9521:
also fixed #3176
image

PR #3232, opened 2026-02-26
manager: fix multi-user app management and add filtering

u9521:
- Implement reflection on `UserManager.getUsers` in `KsuService` to accurately retrieve all user profiles, including work profiles and app clones (dual apps).
- Add a UI dropdown option to filter and display only primary user apps when multiple users are detected.
- Replace `suspendCoroutine` with `suspendCancellableCoroutine` for binding RootService.
- use `RootService.unbind()` for standard lifecycle management of the temporary root service.
- Fix the invalid Mutex instantiation in `SuperUserViewModel`. Replaced it with a class-level `refreshMutex` to correctly queue concurrent refresh requests.
- Update AppProfile route and screens to accept and pass `uid`. Fixed the issue where clicking on a non-primary user's app would redirect to the primary user caused by #3175

PR #3240, opened 2026-03-01
manager: Fix SecurityException when loading icons for secondary users

u9521:
after #3232 we can manage users across different profile groups (e.g., secondary users). However, invoking pm.getUserBadgedIcon for these isolated users triggers a SecurityException due to system permission restrictions on cross-profile interaction.


YuKongA:
我等下有个比较大的更改

PR #3243, opened 2026-03-01
kernel: Use ksu cred to save allowlist

aviraxp:
Some modules mess up allowlist context, which makes init not able to modify it. This workarounds https://github.com/tiann/KernelSU/issues/3234.


sabardarsani-gif (commented):
###

PR #3248, opened 2026-03-03
ksud: set max page size to 16KB for Android targets

u9521:
Fixes ksud execution on Android devices with 16KB page sizes by explicitly setting the max page size during linking.

Solves: 
`
CANNOT LINK EXECUTABLE ".../libksud.so": empty/missing DT_HASH/DT_GNU_HASH in ".../libksud.so" (new hash type from the future?)
`

### DIFF
--- a/userspace/ksud/build.rs
+++ b/userspace/ksud/build.rs
@@ -29,6 +29,9 @@ fn get_git_version() -> Result<(u32, String), std::io::Error> {
 }
 
 fn main() {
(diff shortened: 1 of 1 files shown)


aviraxp:
Does it work now after using ndk29?

5ec1cff:
Superseded by #3251 . [The default alignment of shared libraries for arm64-v86 and x86_64 is 16k since ndk 28](https://github.com/android/ndk/wiki/Changelog-r28), and we are using ndk29 now.

PR #3258, opened 2026-03-08
feat: add late-load mode support

tiann:


kernel/ksu.c
```diff
@@ -46,7 +52,8 @@ int __init kernelsu_init(void)
     canary &= CANARY_MASK;
     __stack_chk_guard = canary;
 #endif
-
+    late_load = (current->pid != 1);
```
aviraxp: Will this break gki mode?

kernel/selinux/rules.c
```diff
@@ -94,6 +96,8 @@ void apply_kernelsu_rules()
     ksu_allow(db, "system_server", KERNEL_SU_DOMAIN, "process", "getpgid");
     ksu_allow(db, "system_server", KERNEL_SU_DOMAIN, "process", "sigkill");
 
+    reset_avc_cache();
```
aviraxp: We are in atomic context in ksu_handle_execveat_ksud, this is likely to cause schedule when atomic. Since it is unlikely for kernel to have explicit deny rules for ksu, i think we should remove it.

PR #3273, opened 2026-03-12
manager: Adjust the order of the theme settings items

WenHao2130:
This commit swaps the menu order of "Theme" and "UI Style" to match the menu order in Miuix.

Change-Id: I9e89815046d79e16e351fc09128856ad1a775e62


KOWX712:
more to-be-changed here https://github.com/KOWX712/KernelSU/commits/ksu_ui/

PR #3275, opened 2026-03-12
manager: Add settings icon to home screen TopBar for uninstalled state

dabao1955:
When KernelSU is not installed, the bottom navigation bar is hidden, making it impossible for users to access Settings to switch between Miuix and Material themes.

Add a settings icon button to the TopBar actions in both HomeMiuix.kt and HomeMaterial.kt, allowing users to access Settings regardless of installation status.

Change-Id: I3018e883a8a97366d85ed2f603a352eae5ac4ac3

### DIFF
--- a/manager/app/src/main/java/me/weishu/kernelsu/ui/MainActivity.kt
+++ b/manager/app/src/main/java/me/weishu/kernelsu/ui/MainActivity.kt
@@ -213,10 +213,10 @@ class MainActivity : ComponentActivity() {
                                 entry<Route.Install> { InstallScreen() }
                                 entry<Route.Flash> { key -> FlashScreen(key.flashIt) }
                                 entry<Route.ExecuteModuleAction> { key -> ExecuteModuleActionScreen(key.moduleId) }
-                                entry<Route.Home> { MainScreen() }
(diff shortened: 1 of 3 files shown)


Ylarod:
I think this is not needed

dabao1955:
> I think this is not needed

Why

dabao1955:
> I think this is not needed

In version 2.1.2 of the kernelsu app, there were settings available when it was not installed, but in the next version, they were simply removed. If you think it is unnecessary to add settings when it is not installed, then you shouldn't have added settings on the uninstalled interface in the first place, rather than telling us now that this change is unnecessary. Moreover, you didn't explain why it is unnecessary.

Ylarod:
1. Settings page is not add by me.
2. I think it's completly not needed to switch theme when not installed, what's your need is to uninstalling the app

dabao1955:
There is a feature in the settings to send logs. If the kernel or app is not working properly, it will also show as not installed. According to your previous logic, if it is unnecessary, then these users also cannot send logs, causing the issues they submit to be closed on the grounds of 'missing necessary information.'

Ylarod:
You cannot get full log without root in this app, so even if you send logs, necessary information is missing

PR #3291, opened 2026-03-14
feat: ksud-managed emulated soft reboot

5ec1cff:


aviraxp:
There are a couple of issues:

1. We have not implemented init rc injection in kernel (and is useless after init stage), so we currently cannot handle the boot-complete stage. A simple workaround is to use resetprop -w to wait for the boot-complete property.

2. At the post-fs-data stage, modules located in modules_update are updated. For modules that include a daemon, this may result in two daemons running simultaneously after the update, so the module itself needs to handle this situation.

3. soft-reboot.sh will give user an impression that it will get executed on every zygote restart, but actually it will only run when triggered via ksud. Consider using another name for clarification.

userspace/ksud/src/init_event.rs
```diff
@@ -151,6 +157,24 @@ pub fn on_boot_completed() {
     run_stage("boot-completed", false);
 }
 
+fn wait_for_boot_completed() -> Result<()> {
```
aviraxp: There is no guarantee that this runs before boot complete (service sh may take long time), so do we have to wait boot complete in another thread?

PR #3292, opened 2026-03-14
ksud: magica: use minify adb_client

5ec1cff:
Before aarch64: 4.1M
After aarch64: 3.1M


Liccsu:
Why delete https://github.com/Kernel-SU/adb_client ???
Now ksud can't be built...

PR #3294, opened 2026-03-15
manager: clean up intent leftover

KOWX712:
- cleanup intent after using action shortcut
- drop name extra from webui intent
- cleanup activity label after using webui
- restore gki warning card color in Material


YuKongA:
还有要改的么

KOWX712:
暂时没了

YuKongA (approved):

PR #3298, opened 2026-03-15
feature: allow shell & force enable adbd

5ec1cff:


YuKongA (approved):

JingMatrix:
The variable used in kernel space `allow_shell` and the variable `allowShell` used in manager UI are causing unnecessary confusion.
Current naming is a bad choice, as it would imply (though not true) that `CONFIG_KSU_DEBUG` forces the patching argument `--allow-shell`.

grio48225-creator (approved):

PR #3299, opened 2026-03-15
Refactor `cred_security_struct`, `find_task_by_vpid` and `strncpy_from_user_nofault`.

shadichy:


kernel/ksud.c
```diff
@@ -63,6 +63,28 @@ static struct work_struct stop_init_rc_hook_work;
 static struct work_struct stop_execve_hook_work;
 static struct work_struct stop_input_hook_work;
 
+long __strncpy_from_user_nofault(char *dst, const void __user *unsafe_addr,
```
hmtheboy154, line 66: better reformat this function, it lost all spacing

shadichy: Yeah right, my bad

aviraxp:
What's the usage of this commit? ksuinit can resolve symbol address on the fly.

shadichy:
> What's the usage of this commit? ksuinit can resolve symbol address on the fly.

It's not just or kernelsu, but also for waydroid. Since probing on non-android linux requires [something similar to ksuinit](https://github.com/shadichy/modloader), and this creates some overhead delays on grepping symbols.

aviraxp:
Are these all the symbols you need for it to load without ksuinit?

shadichy:
> Are these all the symbols you need for it to load without ksuinit?

No of course, not enough. KSU uses a lot of non-exported symbols, most of them are unresolvable. These are just list of resolvable/replaceable ones.

CanerKaraca23:
@shadichy conflicts

Issue #3287, opened 2026-03-14
commit_creds is not safe to be called in atomic context


aviraxp:
### Please check before submitting an issue

- [x] I have searched the issues and haven't found anything relevant
- [x] I will upload bugreport file in KernelSU Manager - Settings - Report log
- [ ] I know how to reproduce the issue which may not be specific to my device

### Describe the bug

1

### Device info

1

PR #3308, opened 2026-03-17
kernel: use syscall table hook to avoid running in atomic context

5ec1cff:
- Replace kprobes in ksud.c with raw syscall table hook
- Handle sucompat and setuid hook with unused syscalls, which is redirected by tracepoint
- Make most of atomic-context-unsafe code running in normal context

Solve https://github.com/tiann/KernelSU/issues/3287


5ec1cff:
Merged in https://github.com/tiann/KernelSU/commit/225ffbbf6bd48e22c6b232a31e705a538664eade

PR #3314, opened 2026-03-21
ksud: magica: Delete ro.boot.selinux and compact after run

aviraxp:
It comes from the cmdline exploit and get set by init. After we have a resetprop with compact support, this can be properly deleted.


aviraxp:
I am not sure if we want to deal with ro.boot.mode.

xmanweb:
> I am not sure if we want to deal with ro.boot.mode.

Do you mean ro.bootmode ?

Issue #2113, opened 2024-10-02
KSU bug in granting root for apps on BlissOS


hmtheboy154:
### Please check before submitting an issue

- [X] I have searched the issues and haven't found anything relevant
- [X] I will upload bugreport file in KernelSU Manager - Settings - Report log
- [X] I know how to reproduce the issue which may not be specific to my device

### Describe the bug

(This issue may look like https://github.com/tiann/KernelSU/issues/2095, but I can't verify yet)
Currently with latest KSU on BlissOS, tested on kernel 6.1 & 6.6 & 6.9 & 6.11, there are some bugs that I started to notice: While some apps like Aurora Store, Neo Store & FoxLog can be able to grant root and use normally, some are broken in functions or just straight up not working:
- Termux can only be able to go to `su` if using `/system/bin/su`, and when in su no $PATH is set other than Termux or KSU own path. 
- MiXplorer can't be able to use `Allow Root` even set with `rootexplorer.root` or `system` template. I tried another app [Amaze File Manager](https://github.com/TeamAmaze/AmazeFileManager) with their `Root Explorer` & `Use legacy listing for root` and it can do things like show the whole `/data` but can't access anything in `/data/app`.
- If using another term app like [TermOnePlus](https://f-droid.org/packages/com.termoneplus/), it will not working even if using either `su` or `/system/bin/su`

All recent BlissOS builds with KernelSU 1.0.1 have this issue. And currently they are on kernel 6.1. I bumped kernel to 6.6 to test locally and it show the same behaviour. For newer kernel that are on stable branch, the same behaviour can be seen on [this PR I made](https://github.com/tiann/KernelSU/pull/1785#issuecomment-2143743204) . Recently I do try to build for kernel 6.11 and it also have the same behaviour (here is [my fork for 6.11](https://github.com/hmtheboy154/KernelSU/tree/6.11)).

### To Reproduce

- For kernel 6.1 + KSU 1.0.1 testing, download any latest stable build of BlissOS
- For kernel 6.1 + KSU 0.9.2 testing, download older builds such as 16.9.6, 15.9.1 or 14.10.2
- For kernel 6.9 + KSU 0.9.2 testing, download latest BlissOS Zenith (16.9.6)
- For kernel 6.6 + KSU 1.0.1 or kernel 6.11 + KSU 1.0.1 tesing, it's currently WIP and still being worked on. If anyone ask, I'll provide.

### Expected behavior

Apps can be able to grant root and work as expected

[...]

PR #3316, opened 2026-03-22
drop x86_64 support

Ylarod:


hmtheboy154:
wait why ?

Ylarod:
> wait why ?

x86_64 breaks new syscall hook method

hmtheboy154:
> > wait why ?
> 
> x86_64 breaks new syscall hook method

But we are still using it here on BlissOS, and there are people who just brought it to WayDroid.

Is it related to this issue ?
https://github.com/tiann/KernelSU/issues/2113

Ylarod:
yes, x86_64 skips sys_call_table

hmtheboy154:
> yes, x86_64 skips sys_call_table

Then I'd suggest bring back the support with a warning. Because this is too sudden though. We can warn until there's a way to fix this.

hmtheboy154:
We also haven't migrated to LKM yet, if that is related

Ylarod:
x86_64 should use GKI mode, I think

hmtheboy154:
> x86_64 should use GKI mode, I think

sure, but please revert and put a warning or something. I don't wanna fork KSU again :cry:

hmtheboy154:
> yes, x86_64 skips sys_call_table

On our kernel, we have a kernel parameter for this. You can use them as a warning.
https://github.com/android-generic/kernel-zenith/commits/6.18_KSU

Ylarod:
> > yes, x86_64 skips sys_call_table
> 
> On our kernel, we have a kernel parameter for this. You can use them as a warning. https://github.com/android-generic/kernel-zenith/commits/6.18_KSU

x86_64 syscall hook infra is also missing

Ylarod:
we can revert changes to arch.h, but syscall_hook.c should be implemented by yourself

hmtheboy154:
> we can revert changes to arch.h, but syscall_hook.c should be implemented by yourself

sure, I'll see if I can do a PR

PR #3319, opened 2026-03-22
kernel: Add ARM64 architecture check

dabao1955:


Ylarod:
some x86_64 maintainers want to implement the x86_64 syscall hook mechanism, so keep the kernel source for easy integration

dabao1955:
> some x86_64 maintainers want to implement the x86_64 syscall hook mechanism, so keep the kernel source for easy integration

If that's the case, then why do you still want to implement 3468172？

Ylarod:
We do not provide official x86_64 anymore, x86_64 maintainers can implement the x86_64 syscall hook mechanism in syscall_hook.h on their own source tree, so they can clone KernelSU as submodule for easy integration. There is not much cost for us.

We do not have much time for maintaining so many devices. Dropping x86_64 support is agreed by most of the core active maintainers.

If you are willing to maintain x86_64 devices, you can fork this repo and do your own changes, then pr the x86_64 support, instead of accusing me why we are dropping the x86_64 support

PR #3320, opened 2026-03-22
ksud: fix the abnormal zygote PID after a soft reboot

ljxi:
`set_next_pid` simulates the behavior of writing to `/proc/sys/kernel/ns_last_pid` in user space

### DIFF
--- a/userspace/ksud/Cargo.lock
+++ b/userspace/ksud/Cargo.lock
@@ -923,6 +923,7 @@ dependencies = [
  "log",
  "nom",
  "prop-rs-android",
+ "rand",
  "regex-lite",
  "rust-embed",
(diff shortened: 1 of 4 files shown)


Ylarod:
This shouldn’t exist on the normal execution path, and the value you hardcoded may not be suitable for all devices. You could write a module that records the zygote PID in the `emulated-soft-reboot` stage, and then complete this kind of handling in the late-load `post-fs-data` stage.

ljxi:
The current implementation cannot control the order in which `post-fs-data.sh` is executed during the post-fs-data phase, so there is no guarantee that other scripts will not consume a large number of PIDs after the module restores them.

PR #3341, opened 2026-03-27
kernel: sucompat: fallback to sh when ksud is unavailable

5ec1cff:


5ec1cff:
https://github.com/tiann/KernelSU/commit/5bf07ba576f094b30b3000dbc8e5b2297fe725e1

PR #3378, opened 2026-03-31
feat: process tag (WIP)

5ec1cff:


backslashxx:
if not for the name tagging, we can just squat on unused bits on thread info flags.

5ec1cff:
> if not for the name tagging, we can just squat on unused bits on thread info flags.

It's for sulog to track the source of root processes

backslashxx:
we can track it easier without any alloc, but we will be squatting on thread info flags

https://github.com/backslashxx/KernelSU/commit/b222853330ba19148b76c7ca1d4c98ebb7c49779
https://github.com/backslashxx/KernelSU/commit/f9b2e86b6058cbe9179e1f1f426050281c60050b

on aarch64, half of thread info flag is available.

https://elixir.bootlin.com/linux/v7.0-rc6/source/arch/arm64/include/asm/thread_info.h

this is long, so wordsize.

5ec1cff:
Process Tag is a mechanism for tracing the origin of root processes, including apps(their uids and package names will be recorded), ksud, modules(their names will be recorded), and manager. We cannot store this information using only a single bit. Furthermore, this is an optional feature; whether to accept the memory overhead is up to the user.

backslashxx:
yeah thats what I was saying

> if not for the name tagging, we can just squat on unused bits on thread info flags.

PR #3387, opened 2026-04-01
fix include problem when building with bazel

aaaaaaaa-815:


5ec1cff:
Did https://github.com/tiann/KernelSU/pull/3390 solve this problem?

aaaaaaaa-815:
> Did #3390 solve this problem?

ok

PR #3390, opened 2026-04-02
kernel: Fix absolute src handling in Kbuild

u9521:
KernelSU's `kernel/Kbuild` used two code paths for include directories:

- `KBUILD_EXTMOD` builds used `-I$(src) -I$(src)/include`
- non-`KBUILD_EXTMOD` builds used `-I$(srctree)/$(src) -I$(srctree)/$(src)/include`

This works for regular kernel builds where `src` is typically a relative path such as `drivers/kernelsu`.

However, Bazel builds pass `src` as an absolute path even when `KBUILD_EXTMOD` is not set. In that case, the non-`KBUILD_EXTMOD` branch produced invalid include paths such as:

- `/workspace/android-kernel/common//workspace/android-kernel/common/drivers/kernelsu`

That broke header resolution for KernelSU internal includes like `uapi/app_profile.h`.

`<asm/memory.h>` is unused in `adb_root.c` and triggers Bazel build failures, so drop the include.

### DIFF
--- a/kernel/feature/adb_root.c
+++ b/kernel/feature/adb_root.c
@@ -1,5 +1,4 @@
 
-#include <asm/memory.h>
 #include <asm/ptrace.h>
 #include <linux/namei.h>
 #include <linux/path.h>


aviraxp:
tg在群里at我一下

PR #3427, opened 2026-04-21
manager: Use Long for versionCode to support values exceeding Int.MAX

Numbersf:
Fixed an issue where versionCode would get stuck at 2147483647. Updated data type to Long to support larger version numbers.


aviraxp:
https://github.com/topjohnwu/Magisk/blob/9d0b5298afd54b320767af7019645f03e96f362f/docs/guides.md?plain=1#L87

PR #3440, opened 2026-05-02
Update what-is-kernelsu.md

santosalexia191ad-a11y:
Quero emoji de iPhone


santosalexia191ad-a11y:
Quero emoji de iPhone

PR #3454, opened 2026-05-08
kernel: allowlist: escape persistent_allow_list to kthread

backslashxx:


backslashxx:
context:

downstream I escape this to root like this
https://github.com/tiann/KernelSU/commit/bf6522cb771cb2706aec8e1b082f587250d98da8
https://github.com/tiann/KernelSU/commit/fcef56aaf44f5ae8481ccea1f17e6bc514ee04d6

but this is probably not needed here as just init cred seems to work.
I've been doing this since ksu v2.

PR #3459, opened 2026-05-11
selinux_hide: fix attr/current detection

5ec1cff:


Ammar64:
When will this be released?

PR #3475, opened 2026-05-18
kernel: Prefer hashed .cfi_jt variants before bare symbols

XiaoTong6666:
On my 5.15.170 device, /proc/kallsyms exposes:
```
  selinux_setprocattr
  selinux_setprocattr.<hash>
  selinux_setprocattr.<hash>.cfi_jt
```

The current functable hook resolver tries the exact ".cfi_jt" name first, but then falls back to the exact bare symbol before walking dotted variants. On this kind of device, that returns the function body address too early, while the LSM hook table stores the hashed ".cfi_jt" address instead, so selinux_hide fails with:
```
  selinux_setprocattr_hook err: -2
```

In the !USE_KCFI path, walk symbol variants before falling back to the exact bare symbol, so hashed CFI jump-table entries are preferred when present.


kernel/infra/symbol_resolver.c
```diff
@@ -157,6 +157,12 @@ void *ksu_resolve_symbol_for_functable_hook(const char *symbol_name)
     addr = (void *)find_kernel_symbol_exact(cfi_name);
     if (addr)
         return addr;
+
+    addr = resolve_symbol_variant(symbol_name, symbol_len);
+    if (addr)
+        return addr;
+
+    return (void *)find_kernel_symbol_exact(symbol_name);
```
5ec1cff, line 165:
```c
    return resolve_symbol_variant(symbol_name, symbol_len);
#else
    addr = (void *)find_kernel_symbol_exact(symbol_name);
    if (addr)
        return addr;

    return resolve_symbol_variant(symbol_name, symbol_len);
#endif
```

5ec1cff (changes requested):

kernel/infra/symbol_resolver.c
XiaoTong6666: 抱歉，打扰您了，小潼有点疑问，就是那这样的话，没有 kallsyms_on_each_symbol() 的老 5.15 的内核上 resolve_symbol_variant() 返回 NULL 了怎么办喵，会不会可能还会像之前那样炸掉 functable_hook() 喵？小潼还是觉得需要留着 exact bare fallback 喵...对不起，小潼不是很懂喵，请波老师指点一下喵...呜呜qwq

5ec1cff: 还真是，那你改成这样

```c
    addr = resolve_symbol_variant(symbol_name, symbol_len);
    if (addr)
        return addr;

    return (void *)find_kernel_symbol_exact(symbol_name);
#else
    addr = (void *)find_kernel_symbol_exact(symbol_name);
    if (addr)
        return addr;

    return resolve_symbol_variant(symbol_name, symbol_len);
#endif
```

XiaoTong6666: 好像当前的 f64ae792aa9198038246bff3870fd115838c6060 就已经是这样的喵？

5ec1cff: 就是说后面那段，因为 `#if` 成立的时候后面完全没用，所以改成 `#else` 的时候生效

XiaoTong6666: 噢噢，懂了，谢谢波老师喵，您看看是这样嘛喵？

5ec1cff: 对的

PR #3521, opened 2026-06-16
kernel, ksud: refactor sucompat

5ec1cff:
- Execute ksud by opening fd with ksu cred and redirecting to execveat, remove READ_SEARCH capability from effective caps.
- Redirect newfstatat/faccessat's `su` path to `ksud` .
- Now sucompat requires `/data/adb/ksud` to exist in caller's root, so chrooted process won't be affected unexpectedly.
- Remove fallback to sh. User should install manager and open it after first installation to install ksud to enable sucompat after first installation.


JingMatrix:
That is very inconvenient design, considering the option of allowing shell su during patching image, which now is useless without manager being opened.

PR #3532, opened 2026-06-22
ksud: Adapt resetprop calls to prop-rs-android 0.2.0

XiaoTong6666:


osm0sis:
This seems like a downgrade to resetprop -c use; no way to blanket apply to all areas anymore without specifying a context for each affected prop?

Why not just improve resetprop -d to automatically compact/rebuild?

XiaoTong6666:
> This seems like a downgrade to resetprop -c use; no way to blanket apply to all areas anymore without specifying a context for each affected prop?

This comes from the `prop-rs-android 0.2.0` API change. The rebuild API now works with a specified context/area:
https://github.com/Kernel-SU/ksu_props/blob/main/crates/prop-rs-android/src/resetprop.rs#L182-L184
https://github.com/Kernel-SU/ksu_props/blob/main/crates/prop-rs-android/src/sys_prop.rs#L729-L737
So this PR only adapts ksud to the new API.

osm0sis:
> This comes from the `prop-rs-android 0.2.0` API change. The rebuild API now works with a specified context/area: https://github.com/Kernel-SU/ksu_props/blob/main/crates/prop-rs-android/src/resetprop.rs#L182-L184 https://github.com/Kernel-SU/ksu_props/blob/main/crates/prop-rs-android/src/sys_prop.rs#L729-L737 So this PR only adapts ksud to the new API.

So you're saying these changes don't affect the CLI resetprop -c? They read like it requires the context area specified now.

XiaoTong6666:
> So you're saying these changes don't affect the CLI resetprop -c? They read like it requires the context area specified now.

Sorry, Sorry, I wasn't clear enough in my previous reply and caused some confusion.😭
Actually, `resetprop -c` is context-based now.
This PR is just syncing ksud with the upstream change. If we want to revisit that behavior, it would probably need to be discussed upstream with @5ec1cff.

osm0sis:
I see. Yeah, what would make the most sense to me from a usability standpoint would be to have every resetprop -d automatically compact the target prop's context area afterwards, and then still have the ability for resetprop -c with nothing specified compact everything.

5ec1cff:
@osm0sis Solved in https://github.com/tiann/KernelSU/pull/3545

nndmt034 (approved):

nndmt034 (approved):

nndmt034 (approved):

PR #3559, opened 2026-07-06
sucompat: optimize string compare

backslashxx:


backslashxx:
https://godbolt.org/z/fY6EaPYq5

PR #3560, opened 2026-07-06
build: allow customize package name and app name when building

5ec1cff:
- ksud: The default package name can be changed by setting the KSU_PACKAGE_NAME environment variable when running cargo.
- manager: The package and application names can be changed by adding the parameters `-PKSU_PACKAGE_NAME=xxx` and `-PKSU_NAME=yyy` to the gradlew command.

### DIFF
--- a/.github/workflows/build-manager.yml
+++ b/.github/workflows/build-manager.yml
@@ -14,6 +14,8 @@ on:
       - 'manager/**'
       - 'kernel/**'
       - 'userspace/**'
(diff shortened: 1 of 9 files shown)


khaidox:
Hi everyone , Does this commit allow fully changing "me.weishu.kernelsu" to a custom package name using only "KSU_PACKAGE_NAME", or are additional kernel-side changes still required?

PR #3572, opened 2026-07-11
kernel: add missing NULL checks on allocation paths

palazik:
Two kernel allocation sites dereference their result without checking for failure, which panics the kernel under memory pressure.

### 1. `kernel/manager/throne_tracker.c` — `my_actor()`

```c
struct apk_path_hash *apk_data = kzalloc(sizeof(struct apk_path_hash), GFP_KERNEL);
apk_data->hash = hash;   // NULL deref if kzalloc fails
```

Every other allocation in this file is checked (e.g. the `data_path` alloc a few lines above skips the entry on failure). This one wasn't. Fixed by skipping the entry (`FILLDIR_ACTOR_CONTINUE`) when the allocation fails.

### 2. `kernel/selinux/sepolicy.c` — `add_filename_trans()`

```c
trans = kcalloc(1, sizeof(*trans), GFP_KERNEL);
struct filename_trans_key *new_key = kzalloc(sizeof(*new_key), GFP_KERNEL);
*new_key = key;                        // NULL deref
new_key->name = kstrdup(key.name, ..); // unchecked
trans->next = last;                    // NULL deref
```

Three unchecked allocations, all dereferenced immediately. Fixed by checking each and bailing out (`return false`, matching the function's existing error style), freeing the partial allocations on the way out.

Both are OOM-only robustness fixes; no behavior change on the success path.

### DIFF
--- a/kernel/manager/throne_tracker.c
+++ b/kernel/manager/throne_tracker.c
@@ -143,6 +143,10 @@ FILLDIR_RETURN_TYPE my_actor(struct dir_context *ctx, const char *name, int name
                 }
             } else {
                 struct apk_path_hash *apk_data = kzalloc(sizeof(struct apk_path_hash), GFP_KERNEL);
+                if (!apk_data) {
+                    pr_err("Failed to allocate apk_path_hash for %s\n", dirpath);
+                    return FILLDIR_ACTOR_CONTINUE;
+                }
                 apk_data->hash = hash;
                 apk_data->exists = true;
                 list_add_tail(&apk_data->list, &apk_path_hash_list);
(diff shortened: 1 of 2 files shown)


juanma0511 (approved):

d15165231-pixel:
Kernel
5.15.189-android13-8-00013-gaa2747a3794a-ab14413887

Versi manager
v0.5.0 (10724)

API Level
36

ABI
arm64-v8a, armeabi-v7a, armeabi

Fingerprint
realme/RMX5313/RE60BA:16/BP2A.250605.031.A3/V.R4T2.1785124421:user/release-keys

Patch keamanan
2026-08-01

Status SElinux
Enforcing

d15165231-pixel:
kernel/manager/throne_tracker.c — my_actor()

PR #3592, opened 2026-07-22
ksud: validate resetprop --timeout to avoid a panic

palazik:
`Duration::from_secs_f64` panics on negative, NaN, or out-of-range input. The `--timeout` value is parsed directly from the command line and passed straight into it:

```rust
let timeout = cli.timeout.map(Duration::from_secs_f64);
```

So something like `resetprop -w --timeout -1 sys.boot_completed` panics instead of printing an error.

Switched to the fallible `try_from_secs_f64` and mapped the error into a regular message, which matches how the rest of the argument handling reports bad input.


5ec1cff:
https://github.com/tiann/KernelSU/commit/9f3c17687035ee760a841a122210ce3dba818841

PR #3597, opened 2026-07-23
fix(kernel): observer init error handling and sucompat error overwrite

EasyGuo114514:
## Summary
- Fix ksu_observer_init() silently ignoring watch_one_dir() failure
- Fix escape_with_root_profile() error overwritten by execveat result

### 1. ksu_observer_init returns error instead of always 0
ksu_observer_init() called watch_one_dir() and stored the return value in ret, but unconditionally returned 0 regardless of the result. Callers in init.c and boot_event.c had no way to detect the failure. Also fixes fsnotify_group leak on the error path.

### 2. escape_with_root_profile error overwritten by execveat result
In ksu_handle_execve_sucompat(), the return value of escape_with_root_profile() was stored in ret, which was immediately overwritten by the execveat result. A scoped local variable now keeps the profile error separate, ensuring sulog captures the correct status.

## Changed files
- kernel/manager/pkg_observer.c
- kernel/feature/sucompat.c

Generated with Claude Code


EasyGuo114514:
fix it

pr_out.txt
```diff
@@ -0,0 +1,1103 @@
+[
```
5ec1cff, line 1: Remove this file

kernel/feature/sucompat.c
```diff
@@ -221,11 +221,13 @@ long ksu_handle_execve_sucompat(const char __user **filename_user, int orig_nr,
...
 
-    ret = escape_with_root_profile();
-    if (ret) {
-        pr_err("escape_with_root_profile failed: %ld\n", ret);
+    {
+        int profile_ret = escape_with_root_profile();
+        if (profile_ret) {
+            pr_err("escape_with_root_profile failed: %d\n", profile_ret);
+        }
+        ksu_sulog_emit_pending(pending_sucompat, profile_ret, GFP_KERNEL);
```
5ec1cff, line 229: This modification is pointless.

5ec1cff (changes requested):

PR #3600, opened 2026-07-25
fix(ui): resolve multiple UI bugs across home, module, template and sulog screens

EasyGuo114514:
## Summary

This PR fixes several UI bugs found during code review of the KernelSU manager interface.

## Details

- **HomeMiuix StatusCard**: Replace hardcoded green colors with theme-derived secondaryContainer/primary so the card follows the active color scheme in both dynamic and non-dynamic color modes.
- **HomeMiuix DonateCard**: Remove inconsistent insideMargin = PaddingValues(18.dp) so it matches LearnMoreCard (default 16dp).
- **HomeMaterial StatusTag**: Use secondaryContainer/onSecondaryContainer instead of errorContainer/onErrorContainer for SafeMode/LateLoadMode tags - these are operational states, not errors.
- **HomeMaterial InfoCard**: Fix asymmetric padding (start=16, end=16, top=14, bottom=12) to uniform 16.dp on all sides.
- **ModuleMaterial and TemplateMaterial**: Fix derivedStateOf anti-pattern where mutable variables were captured and mutated inside the derived block. State is now hoisted into mutableStateOf/mutableIntStateOf/mutableFloatStateOf so mutations are observable across recompositions.
- **SulogMaterial**: Replace selectedEntry!! force-unwrap with safe ?.let call inside the null-check block.

Signed-off-by: EasyGuo114514 <2025520491@qq.com>

### DIFF
--- a/manager/app/src/main/java/me/weishu/kernelsu/ui/screen/home/HomeMaterial.kt
+++ b/manager/app/src/main/java/me/weishu/kernelsu/ui/screen/home/HomeMaterial.kt
@@ -281,16 +281,16 @@ private fun StatusCard(
                             Spacer(Modifier.width(8.dp))
                             StatusTag(
                                 label = stringResource(id = R.string.safe_mode),
-                                contentColor = MaterialTheme.colorScheme.onErrorContainer,
-                                backgroundColor = MaterialTheme.colorScheme.errorContainer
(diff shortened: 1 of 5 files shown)


KOWX712:
?

EasyGuo114514:
？

PR #3609, opened 2026-07-30
feat(ksu): add KSU_IOCTL_DISABLE_KSU to drop KSU capability per process

superturtlee:
Add KSU_IOCTL_DISABLE_KSU (_IO('K', 22), perm only_root) letting a process irrevocably drop all KSU capabilities for itself and its fork/exec children, via a new per-task thread flag TIF_KSU_DISABLE_KSU (bit 62), mirroring the existing TIF_KSU_DISABLE_ESCAPE_WITH_ROOT mechanism.

It disables supercall in addition to TIF_KSU_DISABLE_ESCAPE_WITH_ROOT. (All the ksu features)
It can help solve some container escape bugs of container software without userns(they use seccomp for isolation) such as Droidspaces(https://github.com/ravindu644/Droidspaces-OSS/issues/257)
 

Once set: is_manager() always false (blocks manager-uid impersonation); __ksu_is_allow_uid_for_current() always false (blocks GRANT_ROOT and execve-su); ksu_supercall_handle_ioctl() rejects every ioctl with -EPERM unconditionally; reboot_handler_pre() no longer installs a [ksu_driver] fd; ksu_handle_setresuid() skips fd install / seccomp cache so setresuid to the manager appid cannot auto-install an fd.

The flag is fork-inherited via thread_info.flags memcpy and never cleared, so the restriction is permanent for the whole process tree.

### DIFF
--- a/kernel/hook/setuid_hook.c
+++ b/kernel/hook/setuid_hook.c
@@ -13,6 +13,7 @@
 #include <linux/uidgid.h>
 
 #include "policy/allowlist.h"
+#include "policy/app_profile.h"
 #include "hook/setuid_hook.h"
 #include "klog.h" // IWYU pragma: keep
 #include "manager/manager_identity.h"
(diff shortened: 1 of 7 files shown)


superturtlee:
this will also cause root with low permission(selinux/seccomp) get ksu context and no secomp

PR #3610, opened 2026-07-31
(WIP)

5ec1cff:
import patch from https://github.com/BuSung-dev/Root-My-Galaxy-Payloads/blob/e372abc9bf56000d30149f5998ceea76bd3a4e85/kernelsu/patches/KernelSU-v3.2.5-samsung-kdp-rkp-defex.patch

### DIFF
--- a/kernel/Kbuild
+++ b/kernel/Kbuild
@@ -1,5 +1,8 @@
 kernelsu-objs := core/init.o
 
+kernelsu-objs += compat/samsung_kdp.o
+kernelsu-objs += compat/samsung_defex.o
+
 kernelsu-objs += feature/kernel_umount.o
 kernelsu-objs += feature/sulog.o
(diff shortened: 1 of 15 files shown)


dev-soragoto:
Would it make sense to include polygraphene's Samsung 6.12 follow-up fix 
https://github.com/polygraphene/KernelSU/commit/a5531763971cf034e3f630d31654189a148e5f81 in this PR before it leaves draft?

On Samsung 6.12, kdp_usecount_dec_and_test is no longer available, so the code needs to use kdp_usecount_sub_and_test instead.

I'm currently running the https://github.com/polygraphene/KernelSU/tree/kdp-612 on an h8q / SM-F9710 with kernel 6.12.58, it works for me.

If useful, I can provide build/runtime logs from the device.

Screenshot_20260905_052804_Samsung account

PR #3630, opened 2026-08-11
refactor(kernel): Perform umount for webview zygote

aviraxp:
This was deliberately not done because app cannot run custom code from isolated processes forked from webview zygote. However, hidepid=2 is broken for android again.

In 2020, Jann Horn discovered that when zygote forks and specializes to app zygote, process groups are not cleaned. This makes app zygote and its child have AID_READPROC and bypass hidepid=2. This was marked as CVE-2020-0257 and fixed in 2020-08-01.

However, it is ignored that isolated process can get directly forked from zygote while no app zygote fork is needed. So isolated_app context process still may have AID_READPROC, which provides a way to bypass the hidepid=2 restrictions on /proc.

The impact is limited by SELinux because isolated_app does not have mlstrustedsubject. But it can still read cmdline, comm, maps, mountinfo etc of other isolated processes, mainly webview sandboxed processes.

Bangcle, a famous commercial app protection service in China, started to make use of this recently. An isolated process is forked and iterates through all pids, trying to open their mountinfo.

This is no doubt an exploit (almost same with CVE-2020-0257) and we have already reported it to Google. Before Google fixes it, we have no way but unmount anything for webview zygote. This is expected to break some font modules aimed at webview, but we have no choice.


aviraxp:
Open for discussion. I cannot think of a better idea.

JingMatrix:
To me, it suffice to treat the webview_zygote as a legal target for denylist.
The detection point is on inconsistency, hence I provide `webview_zygote` the same mount namespace that a normal isolated process (belonging to a detector) would receive. See https://github.com/JingMatrix/NeoZygisk/commit/ec29fb101d02324c80ddca7014cc3569b046d487

aviraxp:
As I said in commit message, this is expected to break some font modules aimed at webview, because fonts are usually loaded in sandboxed process forked from webview zygote. Webview zygote is previously not unmounted for this exact reason, as per a user request. So I provided a toggle.

If some fonts work, that is just because a few common fonts are preloaded in zygote.

JingMatrix:
In this case, to address the user need, isn't propagating `webview_zygote` as a configurable process to the profile setting the right move ?

The underlying issue of detection was simply neglecting `webview_zygote` as configurable process.

aviraxp:
The issue is as long as there is one not-unmounted sandboxed isolated process running, the detection will work. So one match = all match. So I don't think it makes too much sense to make it toggable per-app. Also there is no easy way to do this without a userspace daemon, maybe need to parse cmdline in process's stack.

JingMatrix:
I meant show `webview_zygote` as an app which you can configure its profile (deny list).

Not a `per app` sense as you have understood.

PR #3640, opened 2026-08-16
fix(ksud): survive seccomp-blocked reboot() driver probe (#3638)

### LINKED ISSUE #3638: [Bug] init crashes after libksud.so seccomp SIGSYS on OnePlus 13 (OOS 15)

- [x] I have searched the issues and haven't found anything relevant
- [x] I will upload bugreport file in KernelSU Manager - Settings - Report log
- [ ] I know how to reproduce the issue which may not be specific to my device

### Describe the bug

Device & Environment
Device: OnePlus 13 (PJZ110, China)
Android: 15 (OxygenOS 15.0.0.861)
Kernel: 6.6.66-android15-8
KernelSU version: 3.2.5(32525-2)
Root type: temporary root
Selinux: Enforcing
Problem
After soft reboot (need time), the phone reboots itself. logcat -b crash shows:
```
04:52:43.347  me.weishu.kernelsu  SIGABRT  (SecurityException: Writable dex file not allowed)
04:52:44.060  init               SIGABRT  (Fatal signal 6)
04:52:46.394  libksud.so         SIGSYS   (seccomp prevented syscall 142)
```
Key error lines
```
Fatal signal 6 (SIGABRT) in init
Fatal signal 31 (SIGSYS), seccomp prevented syscall 142 in libksud.so
SecurityException: Writable dex file 'main.jar' is not allowed
```
Expected: KernelSU should not crash and should not cause init to crash (which reboots the phone).
Notes: temporary-root setup, cannot modify kernel or seccomp policy. Crash happens after every soft reboot.

### To Reproduce

There are no special steps, just temporary root and soft restart, which may cause this problem later.
I not will upload bugreport file in KernelSU Manager - Settings - Report log
Because there is no root permission after restart.

### Expected behavior

NOT REBOOT

### Screenshots

_No response_

### Logs

logs in file

[crash.txt](https://github.com/user-attachments/files/31090942/crash.txt)

### Device info

Device: OnePlus 13 (PJZ110, China)
Android: 15 (OxygenOS 15.0.0.861)
Kernel: 6.6.66-android15-8

### Additional context

_No response_

JavSaia:
## Summary

Fixes #3638.

On affected devices `ksud` is killed by a fatal `SIGSYS` during the
`reboot()`-based driver probe, which cascades into an init crash and a boot
loop.

## Root cause

`ksucalls::init_driver_fd()` obtains the driver fd by probing:

```
reboot(KSU_INSTALL_MAGIC1, KSU_INSTALL_MAGIC2, 0, &fd)
```

The supercall is backed by a kprobe on `reboot`. seccomp, however, is evaluated
at syscall entry — before the kprobe runs. On a device whose seccomp policy
denies `reboot(2)` in a context KernelSU did not pre-authorize (for example an
unprivileged `su` attempt), the probe raises a fatal `SIGSYS` and aborts the
process. When that process is one init depends on, the abort cascades into an
init crash and a reboot loop.

Reported on OnePlus 13 (OOS 15, kernel `6.6.66-android15-8`), but the mechanism
is not device-specific.

## Fix

Install a narrow `SIGSYS` handler in `ksud` before any code path can issue the
probe. It converts **only** a seccomp-blocked `reboot()` into an `-EPERM`
return, so the caller simply fails to acquire the driver fd and degrades
gracefully instead of being killed. Any other blocked syscall keeps the fatal
default disposition, so genuine seccomp violations are not masked.

The handler is async-signal-safe (no allocation, locking or logging),
classifies the trap via `si_code == SYS_SECCOMP` and the syscall-number
register, and is implemented for both `aarch64` and `x86_64`.

## Testing

`cargo check` and `cargo clippy` (the crate is
`#![deny(clippy::all, clippy::pedantic)]`) pass for both
`aarch64-linux-android` and `x86_64-linux-android` with `-D warnings`.

### DIFF
--- a/userspace/ksud/src/cli.rs
+++ b/userspace/ksud/src/cli.rs
@@ -489,12 +489,85 @@ enum Initrc {
     Refresh,
 }
 
+// Safety net for the reboot()-based driver probe used by `ksucalls`.
+//
+// seccomp is evaluated at syscall entry, before the kernel-side kprobe that
+// backs the supercall can run. On a device whose seccomp policy denies
+// reboot(2) in a context KernelSU did not pre-authorize (for example an
+// unprivileged `su` attempt), the probe in `init_driver_fd` raises a fatal
+// SIGSYS and aborts the process. When that process is one init depends on, the
+// abort cascades into an init crash and a boot loop (#3638).
+//
+// Turn only a seccomp-blocked reboot() into an -EPERM return so the caller just
+// fails to acquire the driver fd and degrades gracefully. Any other blocked
+// syscall keeps the fatal default disposition so genuine violations are not
+// masked.
+#[cfg(any(target_arch = "aarch64", target_arch = "x86_64"))]
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
+extern "C" fn sigsys_handler(
+    _sig: libc::c_int,
+    info: *mut libc::siginfo_t,
+    ctx: *mut libc::c_void,
+) {
+    // Signal context: async-signal-safe operations only (no alloc/lock/log).
+    if info.is_null() || ctx.is_null() {
+        return;
+    }
+
+    unsafe {
+        let uc = ctx.cast::<libc::ucontext_t>();
+
+        // seccomp traps at syscall entry, so the syscall-number register still
+        // holds the blocked number.
+        #[cfg(target_arch = "aarch64")]
+        let nr = (*uc).uc_mcontext.regs[8] as libc::c_long;
+        #[cfg(target_arch = "x86_64")]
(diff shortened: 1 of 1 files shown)


aviraxp:
Discussed before, conclusion is let it crash

aviraxp:
Also your commet is wrong, this is nothing related to init crash. init isn't crash at all, it just ptints a fatal log and is not related to ksu.

Spring-bulid:
okay

MhmRdd:
you can't handle seccomp(2) exceptions where SECCOMP_RET_KILL_PROCESS is set, especially for syscalls like reboot(2) in Android environment.

Issue #3636, opened 2026-08-12
Module repo pages of Hybrid Mount and Magic Mount-rs cause KernelSU to crash


arkzok:
### Please check before submitting an issue

- [x] I have searched the issues and haven't found anything relevant
- [x] I will upload bugreport file in KernelSU Manager - Settings - Report log
- [x] I know how to reproduce the issue which may not be specific to my device

### Describe the bug

When I go into the module repo, tap on Hybrid Mount or Magic Mount-rs and go to the Releases tab, KernelSU crashes. I haven't been able to repeat this with any other modules' pages. I'm assuming it's because both of these have a lot of text in their releases. But what's interesting is that after I visit both the README and the Info tabs, Releases doesn't make it crash anymore. And if I don't visit the Info tab, no matter how much I wait at README, if I go to Releases, it crashes.

### Device info

- Device: Realme GT Neo 3
- OS Version: Android 16
- KernelSU Version: 3.2.5
- Kernel Version: 5.10.226-android12-9-o-g4195cf223d5d

arkzok:
I can't upload the logs, "Failed to upload". I have no idea why, my internet connection is good.

lavenderhorse:
the whole module repo is down

arkzok:
@lavenderhorse Yea, I can see that. It wasn't down at the time this issue was opened.

PR #3641, opened 2026-08-16
manager: defer module repo WebViews until pager settles

Zhanfg:
## Summary

Mitigates a crash reported when navigating from the module README page to
the Releases page for modules with large release descriptions.

The README and release descriptions are rendered using WebViews. During an
animated pager transition, both the previous page and the target page may be
composed at the same time. This can cause the README WebView and multiple
release-description WebViews to coexist temporarily, significantly increasing
memory pressure.

This change defers creation of the heavy Markdown/WebView content until the
pager has settled on the corresponding page.

The same lifecycle handling is applied to both the Material and Miuix module
repository detail implementations.

## Changes

- only compose README Markdown after the pager settles on the README page
- only compose release Markdown after the pager settles on the Releases page
- keep lightweight page structure available during pager animation
- apply the same behavior to Material and Miuix UIs

## Testing

- built KernelSU Manager with:
  `clean assembleRelease -PIS_PR_BUILD=true`
- verified both Manager UI implementations compile successfully

Addresses #3636.

The reported navigation pattern is consistent with overlapping heavy WebView
composition and increased memory pressure. The issue currently has no usable
crash log, so this PR intentionally does not claim that the root cause has been
conclusively proven.

### DIFF
--- a/manager/app/src/main/java/me/weishu/kernelsu/ui/screen/modulerepo/ModuleRepoMaterial.kt
+++ b/manager/app/src/main/java/me/weishu/kernelsu/ui/screen/modulerepo/ModuleRepoMaterial.kt
@@ -443,6 +443,7 @@ fun ModuleRepoDetailScreenMaterial(
                 )
                 when (page) {
                     0 -> ReadmePage(
+                        contentReadyForWebView = pagerState.settledPage == page,
                         readmeHtml = state.readmeHtml,
                         readmeLoaded = state.readmeLoaded,
                         innerPadding = paddedInnerPadding,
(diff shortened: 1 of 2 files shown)


aviraxp:
Have you verified it or this is what AI outputs?

Zhanfg:
I actually ran into this while working on a downstream project based on KernelSU Manager, so I ended up looking into the upstream implementation and fixed it along the way.

In my case the crash happened around README -> Releases navigation on modules with larger release notes. Looking through the code, the pager transition can keep the README WebView alive while the Releases page starts composing additional Markdown/WebViews, which matched what I was seeing downstream.

I did use AI to help with the investigation and implementation, but it wasn't just AI output pasted into a PR — I reviewed the code and built both the Material and Miuix variants myself.

I don't have a clean upstream crash trace proving this is the only root cause, though, so I kept the PR wording as “mitigates” rather than claiming it's conclusively verified.

PR #3647, opened 2026-08-19
selinux hide: spoof status page and avd seqno

chisewaguri:
sets both to the stock value for the kernel: sequence 0 and policyload 0 below 6.10, sequence 4 and policyload 1 on 6.10 and up.

on late load, the backup can be done after policy is modified (such as when we use exploit to turn selinux to permissive then do loadd_policy to reload policy after the exploit fucks it up. so restore avd seqno to 1

### DIFF
--- a/kernel/feature/selinux_hide.c
+++ b/kernel/feature/selinux_hide.c
@@ -317,6 +317,7 @@ static int ksu_selinux_hide_enable()
         pr_err("no backup sepolicy available, please save feature and reboot to retry!\n");
         return -EAGAIN;
     }
+    backup_sepolicy->latest_granting = 1;
     selinux_write_op = find_kernel_symbol_exact("write_op");
     if (!selinux_write_op) {
         pr_err("selinux_hide: no write_op found!\n");
--- a/kernel/feature/selinux_hide.c
+++ b/kernel/feature/selinux_hide.c
@@ -269,12 +269,13 @@ static void initialize_fake_status()
 
     struct selinux_kernel_status *new_status = page_address(new_page);
     memcpy(new_status, status, sizeof(*status));
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 10, 0)
+    new_status->sequence = 4;
+#else
+    new_status->sequence = 0;
+#endif
     if (ksu_late_loaded && !new_status->enforcing) {
-        // In late_load mode, we may be loaded when selinux was set to permissive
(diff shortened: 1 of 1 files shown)


kernel/feature/selinux_hide.c
```diff
@@ -269,12 +269,15 @@ static void initialize_fake_status()
 
     struct selinux_kernel_status *new_status = page_address(new_page);
     memcpy(new_status, status, sizeof(*status));
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 10, 0)
```
aviraxp, line 279: Current code already handles it, that's why we yse backup. We always assume backup is unmodified.

aviraxp (dismissed):

chisewaguri:
thanks for the review. two commits pushed.

changes only kick in when `ksu_late_loaded` now. in that case the loader restores policy with load_policy before insmod, which lands on the status page before my snapshot does, so apps could see non-stock values. we just serve stock end-of-boot values instead: {0,0} below 6.10, {4,1} on 6.10+.

also not touching backup_sepolicy at all anymore. dropped the latest_granting write and the seqno shown to apps comes from a static value instead.

ptal

kernel/feature/selinux_hide.c
chisewaguri: ptal

5ec1cff (approved):

kernel/feature/selinux_hide.c
```diff
@@ -131,6 +131,9 @@ static ssize_t my_write_context(struct file *file, char *buf, size_t size)
     return length;
 }
 
+// stock reads 1; late load may follow a loader load_policy that bumped the backup
+static u32 fake_avd_seqno __read_mostly = 1;
```
aviraxp: I think we don't need an additional definition here

kernel/feature/selinux_hide.c
```diff
@@ -188,6 +191,7 @@ static ssize_t my_write_access(struct file *file, char *buf, size_t size)
     security_compute_av_user(&fake_state, ssid, tsid, tclass, &avd);
 #endif
 
+    avd.seqno = fake_avd_seqno;
     length = scnprintf(buf, SIMPLE_TRANSACTION_LIMIT, "%x %x %x %x %u %x", avd.allowed, 0xffffffff, avd.auditallow,
```
aviraxp, line 193: =1 and move comment here

PR #3650, opened 2026-08-20
Harden runtime paths and module isolation

tryigit:
Fix kernel correctness and performance issues, harden manager tracking, reduce tracing overhead, improve module mount isolation, and fix allowlist migration, umount list bounds, and syscall hook cleanup.


backslashxx:
wow bro you are so smart how to be like you

PR #3665, opened 2026-08-28
kernel: Do not use kallsyms_on_each_match_symbol on x86-64

supechicken:
On x86-64, `kallsyms_on_each_match_symbol` is blocked by IBT on kernel 7.0:

```
Missing ENDBR: kallsyms_on_each_match_symbol+0x0/0x140
```


5ec1cff:
Could you provide the kernel image?

PR #3672, opened 2026-09-02
build: allow pinning KSU version via environment override

FallmoonXD:
## What

Add opt-in environment overrides to pin the KernelSU version instead of deriving it from `git rev-list --count HEAD`:

- **kernel Kbuild**: `KSU_VERSION_OVERRIDE` sets `KSU_VERSION` verbatim
- **ksud build.rs**: `KSU_VERSION_CODE` + `KSU_VERSION_NAME` (both required) replace the git-derived versionCode/versionName

When unset, behavior is completely unchanged.

## Why

Version is currently `30000 + git commit count`. Any downstream fork or out-of-tree build pipeline that adds commits on top of a release tag (e.g. porting vendor-specific patches like Samsung KDP) ends up with a versionCode that no longer matches any official manager release. The manager then shows a spurious "manager version mismatch" warning even though the UAPI is fully compatible.

Concretely: building on top of `v3.3.0` (2601 commits) with 2 extra commits reports `32603` while the official manager is `32601`.

## Example

Pin a CI build to the v3.3.0 release:

```yaml
env:
  KSU_VERSION: 32601
```

```
make ... KSU_VERSION_OVERRIDE=$KSU_VERSION
KSU_VERSION_CODE=32601 KSU_VERSION_NAME=3.3.0 cargo build ...
```

### DIFF
--- a/kernel/Kbuild
+++ b/kernel/Kbuild
@@ -112,6 +112,12 @@ endif
 
 
 # Calculate version if git version is available
+ifdef KSU_VERSION_OVERRIDE
+# Allow pinning the version (e.g. to match an official manager release)
+KSU_VERSION := $(KSU_VERSION_OVERRIDE)
+$(info -- KernelSU version (override): $(KSU_VERSION))
+ccflags-y += -DKSU_VERSION=$(KSU_VERSION)
+else
 ifdef KSU_GIT_VERSION_VALID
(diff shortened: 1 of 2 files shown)


aviraxp:
https://github.com/tiann/KernelSU/pull/3516

PR #3684, opened 2026-09-04
feature: add volume-up fastbootd rescue

ItosEO:
## Summary

- Add an early-boot rescue gesture that records three Volume Up key presses before the input hook is stopped during early boot.
- Expose the rescue state through a new additive supercall and handle it in `ksud post-fs-data` before module scripts run.
- Add a Manager feature switch for enabling or disabling the rescue gesture.

## Behavior

When the gesture is detected, `ksud` asks Android init to reboot with `reboot,fastboot`, entering userspace fastboot on devices with dynamic partitions.

The feature is enabled by default. A missing or unreadable feature configuration keeps the rescue path enabled, so it remains available after `/data` is formatted or when the configuration is damaged. Only an explicitly persisted `0` disables it.

The input hook counts key-down events only. It uses a simple cumulative counter without a time window, matching the existing early-boot Volume Down safe mode approach. The existing three-press Volume Down safe mode behavior remains available, and a failed fastbootd request is logged without interrupting the normal boot flow.

This feature intentionally does not support module management because its decision must be made before module scripts run.

## Compatibility

- The new supercall is additive. `ksud` treats `ENOTTY` from older kernels as no rescue request.
- Fastbootd reboot is requested only when `ro.boot.dynamic_partitions` is enabled.
- The kernel-side feature is registered only for normal built-in initialization; late-loaded KernelSU does not advertise support.

## Validation

- `cargo ndk -t arm64-v8a check`
- `cargo ndk -t arm64-v8a clippy`
- `cargo fmt`
- ClangFormat, Rustfmt, Clippy, Manager, `ksuinit`, all LKM matrices, Android `ksud`, extra desktop targets, and Manager repacking passed in GitHub Actions.


ItosEO:
这个功能的开发原因是，如 OPPO Pad Mini 等设备利用漏洞解锁后，没有办法进入到官方 Fastboot（可以用魔改 EFISP 解除限制），所以我们只能进 fastbootd 来刷分区，但是只能通过系统重启进去，限制比较大，所以加做了这个功能

ItosEO:
@aviraxp @5ec1cff 可以看一下这个 PR 吗？谢谢

PR #3710, opened 2026-09-13
kernel: retain capabilities across execve for non-root profiles

u9521:
Non-root profiles lose capabilities after exec su because non-root processes require ambient capabilities  to inherit privileges.

Populate `cap_inheritable` and `cap_ambient` when `profile->uid != 0` to preserve configured capabilities across execve.


5ec1cff (approved):

akbryrwbyna-tech:
روبینا

PR #3746, opened 2026-09-22
kernel: fix RCU-protected SELinux policy access

XiaoTong6666:
#2646 and #2695 previously fixed invalid RCU usage in "apply_kernelsu_rules()" and "handle_sepolicy()".

After #3268 switched SELinux updates to the copy-and-replace model, both paths started reading "selinux_state.policy" before passing the local pointer to "rcu_dereference_protected()".

This changes them to dereference "selinux_state.policy" directly while holding "policy_mutex", following AOSP common kernel's SELinux policy update implementation:

https://android.googlesource.com/kernel/common/+/refs/heads/android16-6.12/security/selinux/ss/services.c#2226


XiaoTong6666:
~~水一下~~

